PR #29, opened 2023-05-03
Fixed MTASTSHTTPS.NonTLS option

kou029w:
AutoconfigHTTPS.NonTLS option was being used.
Fixed to use MTASTSHTTPS.NonTLS option.


mjl-:
Awesome, thanks for the PR!

PR #33, opened 2023-05-24
Support WebSocket upgrade protocol in reverse proxy

### LINKED ISSUE #25: [Feature] Support WebSocket upgrade protocol in reverse proxy.
I'm trying to install a service (minio) as proxied by mox, and some features do not work because a web socket connection cannot be established.

154pinkchairs:
Resolves #25


http/webserver.go
```diff
@@ -187,7 +191,8 @@ func HandleStatic(h *config.WebStatic, w http.ResponseWriter, r *http.Request) (
 		} else if os.IsPermission(err) {
 			// If we tried opening a directory, we may not have permission to read it, but
 			// still access files inside it (execute bit), such as index.html. So try to serve it.
-			index, err := os.Open(filepath.Join(fspath, "index.html"))
+			var index *os.File
+			index, err = os.Open(filepath.Join(fspath, "index.html"))
```
mjl-, line 195: was this change meant as a way to prevent shadowing the "err" variable?
this one and/or the changes immediately below are causing test failures. the handling of these requests is somewhat tricky.

btw, i'm adding some comments. no expecting you to make changes to this PR at the moment. also see the issue for more info.

http/webserver.go
```diff
@@ -239,15 +244,14 @@ func HandleStatic(h *config.WebStatic, w http.ResponseWriter, r *http.Request) (
...
-			defer index.Close()
-			var ifi os.FileInfo
-			ifi, err = index.Stat()
-			if err == nil {
-				w.Header().Set("Content-Type", "text/html; charset=utf-8")
-				serveFile("index.html", ifi.ModTime(), index)
-				return true
-			}
+		}
+		defer index.Close()
```
mjl-, line 248: i can see index being nil with this change. err != nil is now still an option when we get here.

http/webserver.go
```diff
@@ -409,7 +414,40 @@ func HandleForward(h *config.WebForward, w http.ResponseWriter, r *http.Request,
 		proto = "https"
 	}
 	r.Header["X-Forwarded-Proto"] = []string{proto}
-	// todo: add Forwarded header? is anyone using it?
+	if isWebSocketRequest(r) {
+		// Create a new websocket server
+		websocketHandler := websocket.Handler(func(ws *websocket.Conn) {
```
mjl-, line 419: i think this is the right place to do this. but that we should monitor the websocket handshake ourselves (send it to the backend, read its response, check that it understood this was a websocket upgrade request, and send the response back the client). If the websocket upgrade was successful, we would then just shuffle the bytes between the two endpoints until the connection is closed. I don't think we have to enforce that the framing messages are correct.

mjl-:
the commit for #25 implements the websocket support based on monitoring the http handshake and copying the raw data frames. i took in the idea of using golang.org/x/net/websocket, but only for tests. thanks for the PR, i hope the feedback makes sense!

PR #49, opened 2023-07-22
Add a 'KeepRejects' option that disables auto-cleanup

bobobo1618:
I'd like to retain all mail that comes in, even spam, as I can use it to train spam classifiers in the future.


smtpserver/server.go
```diff
@@ -2418,10 +2418,15 @@ func (c *conn) deliver(ctx context.Context, recvHdrFor func(string) string, msgW
...
 					m.MessageHash = messagehash
 					acc.WithWLock(func() {
-						if hasSpace, err := acc.TidyRejectsMailbox(c.log, conf.RejectsMailbox); err != nil {
-							log.Errorx("tidying rejects mailbox", err)
-						} else if hasSpace {
-							if err := acc.DeliverMailbox(log, conf.RejectsMailbox, m, dataFile, false); err != nil {
+						hasSpace := true
+						var err error = nil
+						if !conf.KeepRejects {
+							if hasSpace, err = acc.TidyRejectsMailbox(c.log, conf.RejectsMailbox); err != nil {
```
mjl-, line 2424: could you move the err != nil check to below outside the "if !conf.KeepRejects" check? so only the TidyRejectsMailbox in the "if !conf.KeepRejects" check. the err-checking logic will stay closer to what it was. otherwise, on error from TidyRejectsMailbox, with the change as is, we will start logging "not storing spammy mail to full rejects mailbox" as well as the error.
looks good otherwise, thanks for the PR!

bobobo1618: Done!

smtpserver/server.go
```diff
@@ -2418,10 +2418,16 @@ func (c *conn) deliver(ctx context.Context, recvHdrFor func(string) string, msgW
...
+						var err error = nil
+						if !conf.KeepRejects {
+							hasSpace, err = acc.TidyRejectsMailbox(c.log, conf.RejectsMailbox)
+						}
+						if err != nil {
 							log.Errorx("tidying rejects mailbox", err)
-						} else if hasSpace {
-							if err := acc.DeliverMailbox(log, conf.RejectsMailbox, m, dataFile, false); err != nil {
+						}
+						if hasSpace {
```
mjl-, line 2429: i was hoping for:

```suggestion
						} else if hasSpace {
```

i don't work with github a lot. this feels overheadful for a these nits, sorry about that. if you know of a better way to do this, i'm interested.

smtpserver/server.go
```diff
@@ -2418,10 +2418,16 @@ func (c *conn) deliver(ctx context.Context, recvHdrFor func(string) string, msgW
...
+							if err = acc.DeliverMailbox(log, conf.RejectsMailbox, m, dataFile, false); err != nil {
```
mjl-, line 2430: in mox i keep the shadowing for err (not other vars). or is there another reason for this change besides reducing shadowing?

```suggestion
							if err := acc.DeliverMailbox(log, conf.RejectsMailbox, m, dataFile, false); err != nil {
```

smtpserver/server.go
bobobo1618: I don't work with Github a lot either so no idea I'm afraid. Done though!

smtpserver/server.go
bobobo1618: I just have no idea what I'm doing with Go. Done!

config/config.go
```diff
@@ -300,6 +300,7 @@ type Account struct {
 		Period time.Duration `sconf-doc:"How long unique values are accepted after generating, e.g. 12h."` // todo: have a reasonable default for this?
 	} `sconf:"optional" sconf-doc:"If configured, messages classified as weakly spam are rejected with instructions to retry delivery, but this time with a signed token added to the subject. During the next delivery attempt, the signed token will bypass the spam filter. Messages with a clear spam signal, such as a known bad reputation, are rejected/delayed without a signed token."`
 	RejectsMailbox     string `sconf:"optional" sconf-doc:"Mail that looks like spam will be rejected, but a copy can be stored temporarily in a mailbox, e.g. Rejects. If mail isn't coming in when you expect, you can look there. The mail still isn't accepted, so the remote mail server may retry (hopefully, if legitimate), or give up (hopefully, if indeed a spammer). Messages are automatically removed from this mailbox, so do not set it to a mailbox that has messages you want to keep."`
+	KeepRejects	bool `sconf:"optional" sconf-doc:"Don't automatically delete mail in the RejectsMailbox listed above. This can be useful, e.g. for future spam training.`
```
mjl-, line 303: ah yes, missing ending double quote. surprising that config/doc.go could have been generated though.

smtpserver/server.go
```diff
@@ -2418,7 +2418,12 @@ func (c *conn) deliver(ctx context.Context, recvHdrFor func(string) string, msgW
 					m.MessageID = messageid
 					m.MessageHash = messagehash
 					acc.WithWLock(func() {
-						if hasSpace, err := acc.TidyRejectsMailbox(c.log, conf.RejectsMailbox); err != nil {
+						hasSpace := true
+						var err error = nil
```
mjl-, line 2422: fyi, the `= nil` isn't needed in go. uninitialized values get the zero value. which is nil for the error type (which is an interface).

bobobo1618: Ah, good point. Done, I think?

config/config.go
bobobo1618: Hopefully fixed

mjl-:
Thanks!

PR #91, opened 2023-11-01
Treat messages as text

mattfbacon:
SMTP messages, which sometimes appear as attachments, are currently treated as "possibly binary" by mox webmail. Since SMTP messages are ASCII, we should be able to treat them as text.


mjl-:
Thanks, makes sense to show content immediately instead of requiring a click.

There is also still a todo in the list at the top of webmail/webmail.ts about just rendering a message inside webmail. In a popup, or simply below the regular email (not sure how we would prevent confusion about what the original message vs the forwarded message is).

Issue #119, opened 2024-01-27
Ability to use mox over bidirectional stream: 2024: mox world’s only functional Go IMAP4rev2 server 


haraldrudell:
With few changes mox imapserver package can be used over a bidirectional socket io.ReadWriteCloser

- temporary or separate file-system storage
- configurable logging
- ability to disable timeouts enabling single-step debugging
- bypassing network stack
- integration-testing against mox IMAP service endpoint instance

mjl-:
Sorry for the late reply, was busy on other work, then at FOSDEM.

Indeed, an IMAP server could be spun up relatively easily on a socket. That's also how the internal tests work. I'm wondering: What kind of use cases do you have in mind? Testing Go client code against the mox server? At least that seems like an understandable and reasonable use-case.

I suppose it could be almost enough to export the `serve` and `limitersInit` functions. mlog.LogModel could be added separately too. Then the imaptemp package could be implemented externally, it feels niche enough for that.

PR #120, opened 2024-01-27
mox IMAP4rev2 service endpoint instance on returned bidirectional stream

haraldrudell:
imaptemp.NewImapTemp runs temporary-storage mox imap-server listening on returned bidirectional stream for integration testing

#119 
Ability to use mox over bidirectional stream: 2024: mox world’s only functional Go IMAP4rev2 server
With few changes mox imapserver package can be used over a bidirectional socket io.ReadWriteCloser
temporary or separate file-system storage
configurable logging
ability to disable timeouts enabling single-step debugging
bypassing network stack
integration-testing against mox IMAP service endpoint instance

### DIFF
--- a/imapserver/internals.go
+++ b/imapserver/internals.go
@@ -1,0 +1,23 @@
+package imapserver
+
+import (
+	"crypto/tls"
+	"net"
+)
+
+// internalsStruct exports some private top-level identifiers
+type internalsStruct struct {
+	// LimitersInit configures mox’ connection rate limiting and max connections
+	LimitersInit func()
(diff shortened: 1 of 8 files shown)


S4tyendra (approved):

mjl-:
@S4tyendra what use case would you have for code like this? would you use it in another project?

S4tyendra:
@mjl-  embedding imap in another go binary ( for sync and local fake mailbox). also, single step debug of imap command handling.
 thats for me, but i approved because the capability is useful, the "idea of a pipe-backed IMAP fixture for tests" (not this package as a stable API)

PR #121, opened 2024-01-31
Improve portability of build scripts

daftaupe:
Not all systems have bash (and maybe sh) in the hardcoded path, so use /usr/bin/env to reach the target.

That allows to build mox on FreeBSD for example.


mjl-:
thanks for the pr!
i'm a bit embarrassed to have hardcoded /bin/bash...

PR #146, opened 2024-03-28
do not require the SMTPUTF8 extension when not needed

### LINKED ISSUE #145: remote smtp server does not implement smtputf8 extension, required by message
mox fails to send mails to some SMTP servers with the following error:
```
mox[6651]: l=error m="storing delivery error" pkg=queue deliveryerror="remote smtp server does not implement smtputf8 extension, required by message, transient, transient" cid=18e66f38ee7 from=robot@infra.example.com attempts=0 cid=18e66f38eec permanent=false code=0 secode=
mox[6651]: l=error m="temporary failure delivering from queue" err="remote smtp server does not implement smtputf8 extension, required by message, transient, transient" pkg=queue msgid=169 recipient=laurent@example.com backoff=7m28s nextattempt=2024-03-24T17:45:05+01:00 cid=18e66f38ee7 from=robot@infra.example.com attempts=0 cid=18e66f38eec permanent=false code=0 secode=
```

The email comes from a Prometheus alert.

Prometheus Alertmanager → mox → recipient SMTP server

mox requires the SMTPUTF8 extension on the recipient SMTP server although none of the email headers contain non-ascii characters. I've checked files in the `data/queue/a/` folder, none of those files contains non-ascii characters. The command `grep -r -P "[\x80-\xFF]" data/queue/` gives no results.

Prometheus uses the golang SMTP Client. This SMTP client uses the SMTPUTF8 extension when the SMTP server supports it (which is the case with mox) even when it's not needed (see: https://pkg.go.dev/net/smtp#Client.Mail).

It seems that mox logs somewhere that the email sent by Prometheus needs the SMTPUTF8 extension, and then requires that the recipient SMTP server also supports this extension. Not sure about that, is this assumption correct?

Could it be possible that mox actively checks if the SMTPUTF8 extension is really mandatory (not rely on the received email), and that it does not requires this extension when it's not needed?

lmeunier:
fix #145

### DIFF
--- a/smtpserver/server.go
+++ b/smtpserver/server.go
@@ -1804,9 +1804,8 @@ func (c *conn) cmdData(p *parser) {
 	}
 
 	// Now that we have all the whole message (envelope + data), we can check if the SMTPUTF8 extension is required.
-	// The check happens only when the client required the SMTPUTF8 extension.
 	var part *message.Part
-	if c.smtputf8 {
+	if c.smtputf8 || c.submission || mox.Pedantic {
 		// Try to parse the message.
 		// Do nothing if something bad happen during Parse and Walk, just keep the current value for c.msgsmtputf8.
 		p, err := message.Parse(c.log.Logger, true, dataFile)
@@ -1821,6 +1820,10 @@ func (c *conn) cmdData(p *parser) {
 		if c.smtputf8 != c.msgsmtputf8 {
 			c.log.Debug("SMTPUTF8 flag changed", slog.Bool("received SMTPUTF8", c.smtputf8), slog.Bool("evaluated SMTPUTF8", c.msgsmtputf8))
 		}
+	}
+	if !c.smtputf8 && c.msgsmtputf8 && mox.Pedantic {
+		metricSubmission.WithLabelValues("missingsmtputf8").Inc()
+		xsmtpUserErrorf(smtp.C550MailboxUnavail, smtp.SeMsg6Other0, "smtputf8 extension is required but was not added to the MAIL command")
 	}
 
 	// Prepare "Received" header.
--- a/smtpserver/server.go
+++ b/smtpserver/server.go
@@ -767,7 +767,7 @@ func command(c *conn) {
 	c.cmd = cmdl
 	c.cmdStart = time.Now()
 
-	p := newParser(args, c.msgsmtputf8, c)
+	p := newParser(args, c.smtputf8, c)
 	fn, ok := commands[cmdl]
 	if !ok {
 		c.cmd = "(unknown)"
@@ -1472,7 +1472,7 @@ func (c *conn) cmdMail(p *parser) {
 	}
 
 	// We now know if we have to parse the address with support for utf8.
-	pp := newParser(rawRevPath, c.msgsmtputf8, c)
+	pp := newParser(rawRevPath, c.smtputf8, c)
 	rpath := pp.xbareReversePath()
 	pp.xempty()
 	pp = nil
--- a/smtpserver/server.go
+++ b/smtpserver/server.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode"
 
 	"golang.org/x/exp/maps"
 	"golang.org/x/text/unicode/norm"
@@ -331,7 +332,7 @@ type conn struct {
 	futureRelease        time.Time // MAIL FROM with HOLDFOR or HOLDUNTIL.
 	futureReleaseRequest string    // For use in DSNs, either "for;" or "until;" plus original value. ../rfc/4865:305
 	has8bitmime          bool      // If MAIL FROM parameter BODY=8BITMIME was sent. Required for SMTPUTF8.
-	smtputf8             bool      // todo future: we should keep track of this per recipient. perhaps only a specific recipient requires smtputf8, e.g. due to a utf8 localpart. we should decide ourselves if the message needs smtputf8, e.g. due to utf8 header values.
+	smtputf8             bool      // todo future: we should keep track of this per recipient. perhaps only a specific recipient requires smtputf8, e.g. due to a utf8 localpart.
 	recipients           []rcptAccount
 }
 
@@ -1897,6 +1898,34 @@ func (c *conn) submit(ctx context.Context, recvHdrFor func(string) string, msgWr
 		metricSubmission.WithLabelValues("badfrom").Inc()
 		c.log.Infox("verifying message From address", err, slog.String("user", c.username), slog.Any("msgfrom", msgFrom))
 		xsmtpUserErrorf(smtp.C550MailboxUnavail, smtp.SePol7DeliveryUnauth1, "must match authenticated user")
+	}
+
+	// Check if the message contains non-ascii characters. If no such characters are found,
+	// the SMTPUTF8 extension is not required.
+	// ../rfc/6531:497
+	isASCII := func(s string) bool {
+		for _, c := range s {
+			if c > unicode.MaxASCII {
+				return false
+			}
+		}
+		return true
+	}
+	c.smtputf8 = !isASCII(c.mailFrom.Localpart.String())
+	for _, rcpt := range c.recipients {
+		if !isASCII(rcpt.rcptTo.Localpart.String()) {
+			c.smtputf8 = true
+			break
+		}
+	}
+	for _, values := range header {
+		for _, value := range values {
+			if !isASCII(value) {
+				c.smtputf8 = true
+				break
+			}
+		}
+
 	}
 
 	// TLS-Required: No header makes us not enforce recipient domain's TLS policy.
--- a/smtpserver/server.go
+++ b/smtpserver/server.go
@@ -1659,7 +1659,7 @@ func (c *conn) cmdRcpt(p *parser) {
 }
 
 // ../rfc/6531:497
-func (c *conn) isSmtpUtf8Required(part *message.Part) bool {
+func (c *conn) isSMTPUTF8Required(part *message.Part) bool {
 	hasNonASCII := func(r io.Reader) bool {
 		br := bufio.NewReader(r)
 		for {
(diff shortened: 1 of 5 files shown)


mjl-:
Thanks, I'll have a closer look tomorrow. First thoughts: Nice, with tests! I'm wondering if "header" has decoded mime q/b-word encoded text into utf-8, which could cause us to use smtputf8 when not needed (it is encoded as ascii).

smtpserver/server.go
```diff
@@ -1899,6 +1900,34 @@ func (c *conn) submit(ctx context.Context, recvHdrFor func(string) string, msgWr
 		xsmtpUserErrorf(smtp.C550MailboxUnavail, smtp.SePol7DeliveryUnauth1, "must match authenticated user")
 	}
 
+	// Check if the message contains non-ascii characters. If no such characters are found,
```
mjl-, line 1903: i think we should do the smtputf8-check earlier. already in cmdData. just above the comment "Prepare "Received" header." (github PR review don't seem to let me make a comment there...). because we start composing headers based on c.smtputf8 from then on.

i think we should also keep a separate flag for whether we use smtputf8 for the message, vs what the incoming transaction has. perhaps we should add a flag "msgutf8" or "outsmtputf8" or something similar, and use that when composing headers and when adding to the queue. the smtputf8 flag is used when parsing commands. setting it to false after a the incoming connection has enabled it could lead to confusing/incorrect errors.

it would also be good to add a log line when we change from smtputf8 to non-utf8.

smtpserver/server.go
```diff
@@ -1899,6 +1900,34 @@ func (c *conn) submit(ctx context.Context, recvHdrFor func(string) string, msgWr
...
+	// ../rfc/6531:497
+	isASCII := func(s string) bool {
+		for _, c := range s {
+			if c > unicode.MaxASCII {
+				return false
+			}
+		}
+		return true
+	}
+	c.smtputf8 = !isASCII(c.mailFrom.Localpart.String())
```
mjl-, line 1914: Localpart.String() marshals the Localpart string type to an encoded string. that won't change whether the string has non-ascii. so we can just check for characters with "isASCII(string(c.mailFrom.Localpart))".

smtpserver/server.go
```diff
@@ -1899,6 +1900,34 @@ func (c *conn) submit(ctx context.Context, recvHdrFor func(string) string, msgWr
...
+	for _, rcpt := range c.recipients {
+		if !isASCII(rcpt.rcptTo.Localpart.String()) {
+			c.smtputf8 = true
+			break
+		}
+	}
+	for _, values := range header {
```
mjl-, line 1921: i've read a bit more into rfc 653[012], which confirms that headers of multiparts can also have utf8, which is also a reason to use smtputf8. so this check is not exhaustive.

for example, see this message by arnt: https://github.com/arnt/eai-test-messages/blob/master/attachment
a smaller message like this would be a good test to add.

the easiest option is probably to just read all headers of all parts, and check for non-ascii. we would need a parsed message.Part. it is currently parsed in message.From() at the beginning of cmdSubmit, but we probably need to refactor a little bit to parse the part earlier, and perhaps just pass it to message.From (we don't want to parse messages multiple times).

something like this will do the check:

```
	hasNonASCII := func(r io.Reader) bool {
		br := bufio.NewReader(r)
		for {
			b, err := br.ReadByte()
			if err == io.EOF {
				break
			}
			xcheckf(err, "read header")
			if b > 0x7f {
				return true
			}
		}
		return false
	}

	var walk func(p *message.Part) bool
	walk = func(p *message.Part) bool {
		if hasNonASCII(p.HeaderReader()) {
			return true
		}
		for _, pp := range p.Parts {
			if walk(&pp) {
				return true
			}
		}
		return false
	}

	fmt.Println("message needs smtputf8:", walk(&part))
```

mjl- (changes requested):
i went through rfc653[012] to page in the requirements. made some suggestions based on that.

smtpserver/server.go
lmeunier: Done.

About the separate `msgsmtputf8` flag, I must confess that I've replaced all occurrences of `c.smtputf8` found in server.go with the new flag without checking much (tests are still all ok, so we must be good...).

smtpserver/server.go
lmeunier: Done.

But I'm not sure the way I did it to avoid parsing the message twice is the best choice.

smtpserver/server_test.go
```diff
@@ -1767,3 +1767,63 @@ func TestFutureRelease(t *testing.T) {
...
+	test(`mjl@mox.example`, `remote@example.org`, "header-ascii", "ascii.txt", false, false, nil)
+	test(`mjl@mox.example`, `remote@example.org`, "header-ascii", "ascii.txt", true, false, nil)
+	test(`mjl@mox.example`, `🙂@example.org`, "header-ascii", "ascii.txt", true, true, nil)
+	test(`mjl@mox.example`, `🙂@example.org`, "header-ascii", "ascii.txt", false, true, &smtpclient.Error{Permanent: true, Code: smtp.C553BadMailbox, Secode: smtp.SeMsg6NonASCIIAddrNotPermitted7})
+	test(`Ω@mox.example`, `remote@example.org`, "header-ascii", "ascii.txt", true, true, nil)
+	test(`Ω@mox.example`, `remote@example.org`, "header-ascii", "ascii.txt", false, true, &smtpclient.Error{Permanent: true, Code: smtp.C550MailboxUnavail, Secode: smtp.SeMsg6NonASCIIAddrNotPermitted7})
+	test(`mjl@mox.example`, `remote@example.org`, "header-utf8-😍", "ascii.txt", true, true, nil)
+	test(`mjl@mox.example`, `remote@example.org`, "header-ascii", "utf8-🫠️.txt", true, true, nil)
+	test(`Ω@mox.example`, `🙂@example.org`, "header-utf8-😍", "utf8-🫠️.txt", true, true, nil)
+	test(`mjl@mox.example`, `remote@idn-🌏️.org`, "header-ascii", "ascii.txt", true, true, nil)
```
lmeunier, line 1828: Not sure about this one.

The utf8 character is only in the recipient domain, not the localpart. If I'm not wrong, the SMTPUTF8 extension is required **only** if the domain is not _translated_ to punycode. So this test should fail?

smtpserver/server.go
mjl-: yes, this is almost entirely fine. but the few calls to newParser should use c.smtputf8.

it's hard to trigger a problem around that: you would have to trigger an error in cmdData after setting msgsmtputf8 to false, then (as client) do another rcptto with a utf8 address. not even sure that would be a valid sequence of commands.

smtpserver/server.go
mjl-: looks good to me!

smtpserver/server_test.go
mjl-: well, the smtpclient will send that literal recipient address to the smtpserver, with the utf8 extension, so smtputf8 and msgsmtputf8 will be true. the rcptto address is also included in the test message in the To header.
i think the test should be for a rcptto address with punycode a-labels (and also use a subdomain under example.org, not a domain that could actually exist, e.g. xn--vg8h.example.org for 🌏️.example.org).

smtpserver/server.go
```diff
@@ -1752,6 +1803,26 @@ func (c *conn) cmdData(p *parser) {
...
+		if err == nil {
+			// Message parsed without error. Keep the result to avoid parsing the message again.
+			part = &p
+			err = part.Walk(c.log.Logger, nil)
+			if err == nil {
+				c.msgsmtputf8 = c.isSmtpUtf8Required(part)
+			}
+		}
+		if c.smtputf8 != c.msgsmtputf8 {
+			c.log.Debug("SMTPUTF8 flag changed", slog.Bool("received SMTPUTF8", c.smtputf8), slog.Bool("evaluated SMTPUTF8", c.msgsmtputf8))
```
mjl-, line 1822: style nit: the attributes shouldn't have spaces. it may even break the logfmt we are generating. we could just keep the names "smtputf8" and "msgsmtputf8". i also try to keep the messages lowercase, though that's not applied consistently.

smtpserver/server_test.go
```diff
@@ -1767,3 +1767,63 @@ func TestFutureRelease(t *testing.T) {
...
+Content-Type: text/plain; charset=UTF-8;
+Content-Disposition: attachment; filename="%s"
+Content-Transfer-Encoding: base64
+
+QW4gYXR0YWNoZWQgdGV4dCBmaWxlLg==
+
+--simple boundary--
+`, mailFrom, rcptTo, headerValue, filename), "\n", "\r\n")
+
+			err := client.Deliver(ctxbg, mailFrom, rcptTo, int64(len(msg)), strings.NewReader(msg), false, clientSmtputf8, false)
```
mjl-, line 1802: nit: the parameter before clientSmtputf8 is eightbitmime, which we may want to set to true for future strictness checks.

mjl- (changes requested):
Awesome, a few more comments, should be ready to merge next.

smtpserver/server.go
```diff
@@ -1654,6 +1658,53 @@ func (c *conn) cmdRcpt(p *parser) {
 	c.bwritecodeline(smtp.C250Completed, smtp.SeAddr1Other0, "now on the list", nil)
 }
 
+// ../rfc/6531:497
+func (c *conn) isSmtpUtf8Required(part *message.Part) bool {
+	hasNonASCII := func(r io.Reader) bool {
```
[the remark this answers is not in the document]
lmeunier: I found two other functions similar to `hasNonASCII`:
* `isASCII` in message/compose.go:169
* `isASCII` in publicsuffix/list.go:162

Maybe we should refactor those functions and declare it only once ? (but where?)

smtpserver/server.go
```diff
@@ -1654,6 +1658,53 @@ func (c *conn) cmdRcpt(p *parser) {
 	c.bwritecodeline(smtp.C250Completed, smtp.SeAddr1Other0, "now on the list", nil)
 }
 
+// ../rfc/6531:497
+func (c *conn) isSmtpUtf8Required(part *message.Part) bool {
```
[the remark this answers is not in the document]
lmeunier: As long as this PR is merged, I'll be happy. :)
Having a PR fully finished, or applying some style changes after merging, both are good for me. So, it's up to you to merge this PR when you decide it's content is worth merging.

smtpserver/server.go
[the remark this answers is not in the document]
mjl-: we could merge it, but not sure if worth the trouble. indeed due to finding a place to put it (i think moxio would be a candidate, it's most like a "misc util" package. also, publicsuffix is what i consider a "leaf package", not great to include a dependency on a misc package. and then there is the required error handling that comes with reading from an io.Reader. since it's a small function, may be better to leave duplicates.

smtpserver/server.go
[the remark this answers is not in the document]
mjl-: i pushed a squashed merge commit just now. not sure if this was the best way to do it. i made a minor tweak afterwards to keep staticcheck happy.

thanks for the fix, much appreciated. looking forward to more. (:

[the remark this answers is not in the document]
lmeunier: About the squashed merge, I'm not sure either it's best way to do it. With github, the workflow is to merge this PR which will create a lot of commits in the main branch. To be honest, I'm not a big fan of the github PR workflow, I personally prefer the gerrit workflow (based on amended commit vs adding new commits) which gives a cleaner git history (only one commit per contribution).

But, anyway, thanks for merging this PR. :)

lmeunier:
So, I guess I can close this PR now.
Closed by https://github.com/mjl-/mox/commit/9c5d2341629f598c298f4b9ae14009d108d814af

PR #152, opened 2024-04-09
queue: retry with another IP when first attempt fails for dualstack remote servers

### LINKED ISSUE #149: Remote SMTP server rejects messages because of mismatch between outgoing ipv6 address and reverse hostname
Some remote SMTP server are rejecting messages sent by mox because the IPv6 address used for outgoing connections does not match the reverse DNS lookup of the same IPv6 address.

```
Apr 05 17:19:13 nixos-rpi mox[96706]: l=error m="permanent failure delivering from queue" err="permanent: 550 5.7.1 Client host rejected: cannot find your reverse hostname, [REDACTED_IPV6]" pkg=queue cid=18eaed64af0 from=robot@infra.example.org attempts=0 permanent=true code=550 secode=7.1 msgid=287 recipient=laurent@example.org
```

My ISP won't let me change the reverse DNS record for my ipv6, so I'm stuck to use only ipv4 for outgoing SMTP connections.

Is it possible the configure mox to use only ipv4 for outgoing SMTP connections? (and still accept incoming connections on ipv4 and ipv6).

lmeunier:
mox was already giving another try for received errors after the `HELO`/`EHLO` command. Now mox do the same for received errors when trying to deliver the message to the remote SMTP server.

This should help to deliver messages to SMTP server that rejects incoming messages because of bad ipv4 or ipv6 configuration (for example for servers checking reverse DNS records). mox will now try to deliver messages on both ip family instead before considering the error as permanent.

fix #149

### DIFF
--- a/queue/direct.go
+++ b/queue/direct.go
@@ -639,13 +639,15 @@ func deliverHost(log mlog.Log, resolver dns.Resolver, dialer smtpclient.Dialer, 
 			err = fmt.Errorf("storing recipient domain tls status: %w", err)
 		}
 	}
-	if err != nil {
+
+	inspectError := func(err error) error {
 		if cerr, ok := err.(smtpclient.Error); ok {
 			// If we are being rejected due to policy reasons on the first
 			// attempt and remote has both IPv4 and IPv6, we'll give it
 			// another try. Our first IP may be in a block list, the address for
 			// the other family perhaps is not.
 			if cerr.Permanent && m0.Attempts == 1 && dualstack && strings.HasPrefix(cerr.Secode, "7.") {
+				log.Debugx("change error type from permanent to transient", err, slog.Any("host", host), slog.Any("secode", cerr.Secode))
 				cerr.Permanent = false
 			}
 			// If server does not implement requiretls, respond with that code. ../rfc/8689:301
@@ -653,9 +655,13 @@ func deliverHost(log mlog.Log, resolver dns.Resolver, dialer smtpclient.Dialer, 
 				cerr.Secode = smtp.SePol7MissingReqTLS30
 				metricRequireTLSUnsupported.WithLabelValues("norequiretls").Inc()
 			}
-			err = cerr
+			return cerr
 		}
-		return deliverResult{err: err}
+		return err
+	}
+
+	if err != nil {
+		return deliverResult{err: inspectError(err)}
 	}
 
 	// SMTP session is ready. Finally try to actually deliver.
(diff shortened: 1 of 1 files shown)


lmeunier:
@mjl- in the end, I reused the retry mechanism already present in the code for errors received at the start of the connexion. No changes in the configuration are needed. I don't know if it's better, or not, compared to https://github.com/mjl-/mox/issues/149#issuecomment-2041007381

I'm also wondering how to write a test for this commit. Any idea will be appreciated.

queue/direct.go
```diff
@@ -639,23 +639,32 @@ func deliverHost(log mlog.Log, resolver dns.Resolver, dialer smtpclient.Dialer,
 			err = fmt.Errorf("storing recipient domain tls status: %w", err)
 		}
 	}
-	if err != nil {
+
+	inspectError := func(err error) error {
+		if err == nil {
```
mjl-, line 644: this can't happen right? seems like only calling it for actual errors is a good invariant, so we can get rid of the check.

queue/direct.go
```diff
@@ -639,23 +639,32 @@ func deliverHost(log mlog.Log, resolver dns.Resolver, dialer smtpclient.Dialer,
...
+			return nil
+		}
 		if cerr, ok := err.(smtpclient.Error); ok {
 			// If we are being rejected due to policy reasons on the first
 			// attempt and remote has both IPv4 and IPv6, we'll give it
 			// another try. Our first IP may be in a block list, the address for
 			// the other family perhaps is not.
 			if cerr.Permanent && m0.Attempts == 1 && dualstack && strings.HasPrefix(cerr.Secode, "7.") {
+				log.Debugx("change error type from permanent to transiant", err, slog.Any("host", host), slog.Any("secode", cerr.Secode))
```
mjl-, line 653: should be "transient".

mjl- (changes requested):
this change also seems useful, or did you run into problems with this? i can imagine that it's better to just not attempt delivering over ipv6 if it can cause trouble. but a mail server can connect dualstack, we can retry for these failures both during EHLO, and when it happens later.

> I'm also wondering how to write a test for this commit. Any idea will be appreciated.

this isn't too easy indeed. i think the approach would be to write a test like those in TestQueue where you create a fake smtp server that expects two connections, that responds with a 5xx 7.x.x on first invocation, and expects another connection attempt which completes the smtp transaction successfully. the smtpclient.DialHook could store which address family was dialed, to check if both ipv4 and ipv6 was tried. a mock resolver that resolves to both ipv4 and ipv6 would be needed.
we currently also don't have a check if the retry. would be good to have. the TestQueue code is a bit read-only... if you want to give a try, it may be easier to copy some of that function into a new one: the various cases are connected, and they may require quite a refactor to fit another new case in...

lmeunier:
> this change also seems useful, or did you run into problems with this?

I didn't encounter any problems during my tests with this PR. But, I would prefer to use the other PR for these reasons:
* mails refused over ipv6 are delayed like any failed delivery with transient error (and i don't like to wait ^^): mox first try with ipv6, `RCPT` command is rejected, then mox try agin a few minutes later with ipv4, and finaly the mail is delivered
* I fear that repeated errors could result in my ipv6 being blacklisted in some RBL, and that my ipv6 will be rejected by other SMTP servers
* this PR is a lot harder to test and is less predictable (TBH i'm not sure i haven't break something else with this PR...)

mjl-:
this change didn't look too dangerous to me wrt changing existing behaviour. and it will be useful to also detect this error code during smtp rcpt to in case of dns block listings of ipv6 vs ipv4, so i merged it, thanks!

PR #153, opened 2024-04-11
add TransportDirect transport

### LINKED ISSUE #149: Remote SMTP server rejects messages because of mismatch between outgoing ipv6 address and reverse hostname

lmeunier:
The `TransportDirect` transport allows to tweak outgoing SMTP connections to remote servers. Currently, it only allows to select network IP family (ipv4, ipv6 or both).

For example, to disable ipv6 for all outgoing SMTP connections:
- add these lines in mox.conf to create a new transport named "disableipv6":
```
Transports:
  disableipv6:
    Direct:
      DisableIPv6: true
```
- then add these lines in domains.conf to use this transport:
```
Routes:
  -
    Transport: disableipv6
```

fix #149


lmeunier:
@mjl- this is my second attempt to fix #149. This time by adding a new transport for direct connexions as we talked https://github.com/mjl-/mox/issues/149#issuecomment-2041007381

config/config.go
```diff
@@ -262,6 +263,13 @@ type TransportSocks struct {
 	Hostname dns.Domain `sconf:"-" json:"-"` // Parsed form of RemoteHostname
 }
 
+type TransportDirect struct {
+	DisableIpv4 bool `sconf:"optional" sconf-doc:"If set, outgoing SMTP connexions will *NOT* use IPv4 addresses to connect to remote SMTP servers."`
```
mjl-, line 267: we should name these DisableIPv4 and DisableIPv6 stay in style.

mox-/config.go
```diff
@@ -947,6 +960,9 @@ func PrepareStaticConfig(ctx context.Context, log mlog.Log, configFile string, c
 			n++
 			checkTransportSocks(name, t.Socks)
 		}
+		if t.Direct != nil {
+			checkTransportDirect(name, t.Direct)
```
mjl-, line 964: an "n++" is missing here.

queue/queue.go
```diff
@@ -1075,7 +1075,7 @@ func deliver(log mlog.Log, resolver dns.Resolver, m Msg) {
 			}
 			ourHostname = transport.Socks.Hostname
 		}
-		recipientDomainResult, hostResults = deliverDirect(qlog, resolver, dialer, ourHostname, transportName, msgs, backoff)
+		recipientDomainResult, hostResults = deliverDirect(qlog, resolver, dialer, ourHostname, transportName, transport, msgs, backoff)
```
mjl-, line 1078: shouldn't we be passing transport.Direct here? it can just be nil. it will be clearer in the definition of deliverDirect() what can be expected of the transport (only a *config.TransportDirect).

mjl- (changes requested):
this seems like a nice way to solve the issue. i also saw that using GatherIPs to filter the IPs returns the expected value of dualstack (for the retries around issues with ipv4/ipv6). and gatherips is relatively easy to test. no open/known issues with this, right?

PR #187, opened 2024-06-29
Set syslog facility to mail

kiekerjan:
Setting syslog facility to mail will redirect log output to `/var/log/mail.log` instead of the `/var/log/syslog` file

### DIFF
--- a/mox.service
+++ b/mox.service
@@ -13,6 +13,7 @@ ExecStart=/home/mox/mox serve
 RestartSec=5s
 Restart=always
 ExecStop=/home/mox/mox stop
+SyslogFacility=mail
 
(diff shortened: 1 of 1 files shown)


mjl-:
hi @kiekerjan, this looks fine to me, thanks!
with this setting, mail logging indeed goes to /var/log/mail.log. with all log levels, but that's expected mox doesn't do explicit syslog and syslog doesn't know the levels.
logging still also goes to /var/log/syslog on my machines. i may have changed the logging settings, or it may be the default. admins can change such settings themselves.

kiekerjan:
You're right of course, where the log ends up depends on the configuration of syslog. This only tells syslog that the log facility is mail.

PR #252, opened 2024-11-23
Split paste into address fields by commas

mattfbacon:
If the user has a list of addresses in the clipboard and pastes it in, split the input to have one field for each comma-separated entry. This is more user-friendly.

I don't know if there are cases where a comma in the field is necessary, but if so we could possibly add a workaround by holding a key, e.g. Ctrl+Shift+V for "raw paste" without the splitting behaviour.

### DIFF
--- a/webmail/webmail.js
+++ b/webmail/webmail.js
@@ -2862,6 +2862,23 @@ const compose = (opts, listMailboxes) => {
 		}, function change() {
 			autosizeElem.dataset.value = inputElem.value;
 			fetchRecipientSecurity();
+		}, function paste(e) {
+			const data = e.clipboardData?.getData('text/plain');
+			if (typeof data !== 'string' || data === '') {
+				return;
+			}
+			const split = data.split(',');
+			if (split.length <= 1) {
+				return;
+			}
+			autosizeElem.dataset.value = inputElem.value = split[0];
+			let last;
+			for (const rest of split.splice(1)) {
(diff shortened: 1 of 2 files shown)


mattfbacon:
PS we could also make it so in general if you type "," (or ", ") in the address field it creates a new one for you.

mjl-:
This is great, thanks!
I needed this a few days ago to send to a longish list of addresses, but because it wasn't easy enough to do with the webmail, I sent with Thunderbird.

Comma-handling could indeed be better. Some people use "Lastname, Firstname" as their display name. You could end up pasting `"Lastname, Firstname" <lastname@example.org>`, which would now get split up unintentionally. It would be good to fix that, but pasting multiple comma-separated addresses will be more common, so I consider the new behaviour more helpful. To display names and addresses more completely, we could perhaps add an API method that uses existing Go code for parsing message headers containing addresses. I think it will recognize display names with double quotes and so on, and return a parsed list of addresses.

I'm also liking the idea of ',' to add another field. I don't think there's a need to ever write a literal comma in that field. Thunderbird also handles ',' to finish the address and start a new one.

mattfbacon:
Glad you like it. To be clear, are you planning to now work on what you described?

Btw the type of rules you described are fairly easy to implement in JS I think. I would be concerned about latency and other issues because I already have the server connection drop out relatively regularly while I'm composing a message.

mattfbacon:
Could also be done as paste, send entire pasted string to API for security bar info, API sends info for all addresses back nicely split for us, and we modify the UI accordingly.

PR #287, opened 2025-02-14
Skip 8bitmime check in non-pedantic mode

mattanja:


mattanja:
This is my first contribution to mox and my first line of code in Go - so please review carefully. I have tested sending a mail that was not delivered before, however I don't know if any unit tests need to be updated or added.

The idea about excluding the check in non-pedantic mode came from this line:
https://github.com/mjl-/mox/blob/93b627ceab9b90cb3669d0adc5d6dc208181fc7a/smtpserver/server.go#L2149

mjl-:
@mattanja Thanks for reporting and for the diff. The commit above has a slightly different approach: Only the smtp delivery path from the queue is now skipping the check for the 8bitmime extension. The initial motivation was that I don't want smtpclient to import github.com/mjl-/mox/mox-, since it is a reusable package (listed in apidiff/packages.txt), and the "mox-" package imports too many other packages (like the database library) and mox internals. Package queue isn't reusable, and already imports "mox-". The side-effect is that skipping the check is now only done for delivery (good for public interoperability), not for submission.

Issue #194, opened 2024-07-26
Running mox unprivileged with CAP_NET_BIND_SERVICE


tzvetkoff:
Basically, by adding
```
AmbientCapabilities=CAP_NET_BIND_SERVICE
```
to `mox.service` one can bind privileged ports on Linux (and FreeBSD?)

My understanding is that the only thing that _really_ requires running as the privileged user is binding ports `<1024`.
File access is not (and should not be) of any concern.

This is more of a question than an actual issue, since I'd like to hear your input on a proper way to move this forward.
Here's my attempt to get rid of the _angel_ process - https://github.com/mjl-/mox/compare/main...tzvetkoff:mox:unpriv?expand=1.
It's working totally fine on my server.

mjl-:
Hi @tzvetkoff, that diff indeed seems like a good approach. You're right that starting as root is currently only for being to bind to the privileged ports.
Perhaps in the future mox may want to do more sandboxing. When that happens, it will probably be done with OS-specific mechanisms/APIs. We'll have to see when/if we get there.

A few remarks:
-  The code has a few checks for Getuid() == 0, to guard printing that we're listening on sockets. One example is at https://github.com/mjl-/mox/blob/v0.0.11/imapserver/server.go#L347. I think you're losing those prints now. Would be good to still print them, but still not in localserve mode.
- The name of flag "-skip-forkexec" sounds a bit low-level. A name that explains its purpose at a higher level may be better, e.g "-no-root" or "-unprivileged" (just thinking out loud). With a bit more explanation in the flag or the help text of the command.

tzvetkoff:
Well, this _issue_ was more like a question initially, and most of it can be disregarded when I convert it to a PR (in the next few days, I hope.)

Thing is, I just did a very quick hack just to _show_ we don't need UID-0 to go forward.

Moreover, one can easily `syscall.Setuid(uid)` on UNIX, so the whole lot of shenanigans currently done to circumvent unprivileged execution are just needless.

I totally agree with you that my _diff_ is substandard, it was just to prove the point that `mox` can (and, IMHO, should) use OS-based isolation, and just that if needs be - do proper `syscall.Setuid` after opening the listening sockets.

I will, in a matter of a few days, convert this into a fully-featured pull request, and I'll close this as the conversation will continue there.

Once again, thanks for the great project, and I hope I meet you next year at FOSDEM (I kinda missed it this year.)

mjl-:
[...]

PR #320, opened 2025-03-25
Allow `mox serve` to run unprivileged

tzvetkoff:
Well, I converted this to a pull request since I guess we can think a little bit more on going forward with this.

As discussed and mentioned by @mjl- in #194, I don't see a clear path going forward with `CAP_NET_BIND_SERVICE`, so I'm dropping this one (even though at least FreeBSD provides a way to do that using `sysctl` but I find it insecure since it allows privileged ports *to everyone*.)

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -19,7 +19,7 @@ any parameters. Followed by the help and usage information for each command.
 # Usage
 
 	mox [-config config/mox.conf] [-pedantic] ...
-	mox serve
+	mox serve [-unprivileged]
 	mox quickstart [-skipdial] [-existing-webserver] [-hostname host] user@domain [user | uid]
 	mox stop
 	mox setaccountpassword account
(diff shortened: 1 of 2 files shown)


thenktor:
> (even though at least FreeBSD provides a way to do that using `sysctl` but I find it insecure since it allows privileged ports _to everyone_.)

In FreeBSD you would do it that way. Example from caddy post-install script:
```txt
Note while Caddy currently defaults to running as root:wheel, it is strongly
recommended to run the server as an unprivileged user, such as www:www --

- Use security/portacl-rc to enable privileged port binding:

  # pkg install security/portacl-rc
  # sysrc portacl_users+=www
  # sysrc portacl_user_www_tcp="http https"
  # sysrc portacl_user_www_udp="https"
  # service portacl enable
  # service portacl start

- Configure caddy to run as www:www

  # sysrc caddy_user=www caddy_group=www
```

The last one is an option for the caddy rc (service startup) script, that then will start caddy with specified user. The other stuff should be self explaining: allow a user `www` to bind ports "http https".

PR #332, opened 2025-04-09
Set email address inputs to type email for web-account and -mail

omartijn:
This makes it easier to log in on devices using an on-screen keyboard, as they will change the keyboard layout for easier input of email addresses.

### DIFF
--- a/webaccount/account.ts
+++ b/webaccount/account.ts
@@ -61,6 +61,7 @@ const login = async (reason: string) => {
 								dom.div('Email address', style({marginBottom: '.5ex'})),
 								autosize=dom.span(dom._class('autosize'),
 									username=dom.input(
+										attr.type('email'),
 										attr.required(''),
 										attr.autocomplete('username'),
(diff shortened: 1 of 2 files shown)


mjl-:
Hi @omartijn, thanks for creating the PR. This looks like an obvious improvement, but I don't think it's the correct change: 

1. In the webmail, the address value can essentially be a From-header value. So not only `user@domain`, but also `Firstname Lastname <user@domain>`.  The validation done by an input of type email would reject the second form.
2. Validation of an input with type email rejects internationalized addresses, so we can't use it in practice. See https://github.com/whatwg/html/issues/4562 for details.

What could perhaps work, is using `autocomplete=email`, https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#email. I haven't tried it, so I don't know if it influences how on-screen keyboards are displayed. Perhaps you could give it a try and see if it works? The webmail input field already has an autocomplete attribute btw, you could have to change the existing field.

PR #342, opened 2025-04-25
fix typos in referrer header

wheresalice:
It should be referer in both cases, which whilst misspelled is the actual header name


mjl-:
@wheresalice thanks!

PR #351, opened 2025-05-08
importmaildir: sort messages by date first

### LINKED ISSUE #341: Maildir import seems to mess up order
After importing a single maildir message order seems random in Snappymail.

When migrating from old machines I collected all my emails in a single maildir, removed duplicates and fix mtime (on ext4), on disk this looks reasonably well.

After importing this to mix things still look fine in mox's own webmail, as far as I can tell mox sorts messages by `received` then.

Snappymail OTOH apparently depends on SORT, which isn't yet available, and shows messages pretty much randomly sorted.

I think this is due to `ReadDir()` in `MaildirReader` using "directory order", that appears to be something else than ascending-in-mtime, at least in my case.

I think it might help using other tools (mutt, python,...) to convert to mbox before importing but yet have to give it a shot.

diresi:
maildir messages usually have timestamps in their filename indicating
the `received` date, if that's not the case mtime is used.

since we cannot assume the order of `ReadDir()` entries we have to read
all entries and sort them accordingly first.

fixes #341

### DIFF
--- a/store/import.go
+++ b/store/import.go
@@ -202,14 +202,47 @@ func (mr *MboxReader) Next() (*Message, *os.File, string, error) {
 	return m, mf, mr.Position(), nil
 }
 
+type FileInfo struct {
+	path string
+	time time.Time
+}
+
 type MaildirReader struct {
 	log          mlog.Log
 	createTemp   func(log mlog.Log, pattern string) (*os.File, error)
 	newf, curf   *os.File
-	f            *os.File // File we are currently reading from. We first read newf, then curf.
-	dir          string   // Name of directory for f. Can be empty on first call.
-	entries      []os.DirEntry
+	files        []FileInfo
 	dovecotFlags []string // Lower-case flags/keywords.
+}
+
+// Take received time from filename, falling back to mtime for maildirs
+// reconstructed some other sources of message files.
+func messageTime(f os.DirEntry) time.Time {
+	var t time.Time
+	parts := strings.SplitN(filepath.Base(f.Name()), ".", 3)
+	if v, err := strconv.ParseInt(parts[0], 10, 64); len(parts) == 3 && err == nil {
+		t = time.Unix(v, 0)
+	} else if fi, err := f.Info(); err == nil {
+		t = fi.ModTime()
+	}
+	return t
+}
+
+func readDir(dir *os.File) ([]FileInfo, error) {
+	dn := dir.Name()
+	entries, err := dir.ReadDir(0)
+	if err != nil {
+		return nil, err
+	}
+	var files []FileInfo
+	for _, e := range entries {
+		info := FileInfo{
+			path: filepath.Join(dn, e.Name()),
+			time: messageTime(e),
+		}
+		files = append(files, info)
+	}
+	return files, nil
 }
 
 func NewMaildirReader(log mlog.Log, createTemp func(log mlog.Log, pattern string) (*os.File, error), newf, curf *os.File) *MaildirReader {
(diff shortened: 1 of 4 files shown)


diresi:
How about this? Does this look meaningful? And Golang-ish enough?

store/import.go
```diff
@@ -202,23 +202,55 @@ func (mr *MboxReader) Next() (*Message, *os.File, string, error) {
...
+func readDir(dir *os.File) ([]FileInfo, error) {
+	dn := dir.Name()
+	entries, err := dir.ReadDir(0)
+	if err != nil {
+		return nil, err
+	}
+	var files []FileInfo
+	for _, e := range entries {
+		info := FileInfo{
+			path: filepath.Join(dn, e.Name()),
```
mjl-, line 240: a good part of the reason we reading the entries in the maildir in batches instead of loading them all at once was to reduce memory use. so below some comments about reducing memory use.

could you keep the filename (e.Name()) as separate (original) string, and only join it with the directory name when the file is opened? for large maildirs (cur directories), that will save a whole bunch of allocations.

we can just process "cur" first (sorted), then "new" (sorted separately). "new" contains messages not yet processed by mail clients, so they should all be newer than all messages in "cur", and i believe "new" is often empty (with an active mail client).

when processing cur & new separately, FileInfo doesn't need to get an additional field to indicate whether it's from cur or new.

the entries in "cur", and then "new", can also be read when first needed by Next(). that way, you can easily return an error. the current version of the MR logs an error when a directory can't be read, but doesn't abort and return the error. an admin could easily miss the message, think everything was imported, but the user would miss messages later on.

i think it also makes sense for readDir() to read the directory with multiple dir.ReadDir calls. with dir.ReadDir(0), it will read all dir entries in memory, then process them one by one, by building a new slice with all files in them.

i think it also makes sense to store the time in FileInfo as "nsec int64" instead of time.Time (which has 2 int64's and a pointer).

store/import.go
```diff
@@ -230,43 +262,42 @@ func NewMaildirReader(log mlog.Log, createTemp func(log mlog.Log, pattern string
...
 	}
 
+	files, err := readDir(mr.newf)
+	log.Check(err, "reading new/ directory")
+	files1, err := readDir(mr.curf)
+	log.Check(err, "reading cur/ directory")
+
+	mr.files = append(files, files1...)
+
+	if len(mr.files) == 0 {
```
mjl-, line 272: we don't need a special case of 0 files. we can just sort it, it will be done quickly. in general it's good to have fewer branches/separate code paths, for testability.

store/import.go
```diff
@@ -202,23 +202,55 @@ func (mr *MboxReader) Next() (*Message, *os.File, string, error) {
...
-	entries      []os.DirEntry
+	files        []FileInfo
 	dovecotFlags []string // Lower-case flags/keywords.
 }
 
+// Take received time from filename, falling back to mtime for maildirs
+// reconstructed some other sources of message files.
+func messageTime(f os.DirEntry) time.Time {
+	var t time.Time
+	parts := strings.SplitN(filepath.Base(f.Name()), ".", 3)
```
mjl-, line 222: i think think we need to call filepath.Base on the name. i know we did this previously too, but these entries are just read straight from a directory listing, the strings are just the bare file names (if i'm mistaken, the tests should catch it). the same should apply for the other call to filepath.Base.

mjl- (changes requested):
thanks for the PR, i would like to get this in! (sorry for the very long delay, thanks for sticking with it).

i have some suggestions for changes. if you could make them, great. if you would prefer i make them, that's fine by me too.

diresi:
Hi,

I'm happy you like it so far!
I'd prefer you make your changes, though. It's been a while and I don't feel up to the task anymore, sorry!

Thanks for everything!

mjl-:
I completed this with https://github.com/mjl-/mox/commit/444615004b0ea990481af0b8dc1a01ebb674da4a, thanks again for code & issue!

PR #367, opened 2025-06-23
use ClientSettingsDomain for RFC6186 SRV records if set

### LINKED ISSUE #366: ClientSettingsDomain ignored in RFC6186 SRV records
After configuring the `example.com` domain, `admin/#domains/example.com/dnsrecords` recommends the following [RFC6186](https://www.rfc-editor.org/rfc/rfc6186#section-3) DNS records, among others:

```
; For secure IMAP and submission autoconfig, point to mail host.
_imaps._tcp.example.com.        SRV 0 1 993 mx.example.com.
_submissions._tcp.example.com.  SRV 0 1 465 mx.example.com.
```

Since I've configured `mail.example.com` as the `ClientSettingsDomain`, shouldn't the record point to that instead of `mx.example.com`? E.g.:

```
; For secure IMAP and submission autoconfig, point to mail host.
_imaps._tcp.example.com.        SRV 0 1 993 mail.example.com.
_submissions._tcp.example.com.  SRV 0 1 465 mail.example.com.
```

I’ve tested `autoconfig` and `autodiscover` with `curl`, and they correctly return `mail.example.com`.

allddd:
make the DNS record recommendations consistent with autoconfig/autodiscover and make it easier to migrate to a different server if clients that implement autodiscovery as described in RFC6186 are used

Closes #366.

### DIFF
--- a/webadmin/admin.go
+++ b/webadmin/admin.go
@@ -1402,10 +1402,13 @@ When enabling MTA-STS, or updating a policy, always update the policy first (thr
 			}
 		}
 		srvhost := func(ok bool) string {
-			if ok {
-				return mox.Conf.Static.HostnameDomain.ASCII + "."
+			if !ok {
+				return "."
 			}
-			return "."
+			if domConf.ClientSettingsDomain != "" {
+				return domConf.ClientSettingsDNSDomain.ASCII + "."
+			}
+			return mox.Conf.Static.HostnameDomain.ASCII + "."
 		}
 		var reqs = []srvReq{
 			{name: "_submissions", port: 465, host: srvhost(submissions)},


mjl-:
Hi allddd!

This appears to make sense, but I had a vague memory about why it isn't implemented like this already: The target of a SRV record cannot be an alias, but must be an address record. See https://datatracker.ietf.org/doc/html/rfc2782 on page 3:

```
The format of the SRV RR

   Here is the format of the SRV RR, whose DNS type code is 33:

        _Service._Proto.Name TTL Class SRV Priority Weight Port Target

[...]

Target
        The domain name of the target host.  There MUST be one or more
        address records for this name, the name MUST NOT be an alias (in
        the sense of [RFC 1034](https://datatracker.ietf.org/doc/html/rfc1034) or [RFC 2181](https://datatracker.ietf.org/doc/html/rfc2181)). 
```

The first page of the rfc has this:

```
   Note that where this document refers to "address records", it means A
   RR's, AAAA RR's, or their most modern equivalent.
```

I believe the "alias" from 1034 refers to the dns name with a CNAME record.

I don't know the rationale for disallowing a cname as the srv target, or what clients are supposed to do if the name is a cname. I suspect most clients wouldn't even notice. But I prefer not to violate the standards, even though the client domain settings cname would seem useful in this case.

mjl-:
@allddd I rebased your commits and merged them, along with an updated selfcheck that won't complain when the previous static hostname is still configured for the target in the srv record. I also added a reference to rfc 2782 and arnts explanation. Thanks all! Closing this now, feel free to reopen if more changes are needed.

Issue #369, opened 2025-07-02
No real ip behind haproxy 0.0.15


vooon:
Hello,

I'm trying to setup Mox behind the HAproxy, UI/MTA-STS/... works, but seems Mox not processing `X-Forwarded-For` / `X-Real-IP` because in logs and account client ip is `127.0.0.1`.

`mox.conf` is like that:
```
Listeners:
	internal:
		AccountHTTP:
			Enabled: true
			Port: 1080
			Forwarded: true
		AdminHTTP:
			Enabled: true
			Port: 1080
			Forwarded: true
		WebmailHTTP:
			Enabled: true
			Port: 1080
			Forwarded: true
		WebAPIHTTP:
			Enabled: true
			Port: 1080
			Forwarded: true
```

`haproxy.cfg`:
```
frontend https
  mode http
  bind :443 ssl alpn h2,http/1.1 crt /etc/acme/haproxy/
  # bind quic4@:443 ssl alpn h3 crt /etc/acme/haproxy/
  # http-after-response add-header alt-svc 'h3=":443"; ma=60'
  option forwardfor
  http-request set-header X-Real-IP %[src]
  http-request set-header X-Forwarded-Proto https
  http-response set-header Strict-Transport-Security "max-age=2592000"

  use_backend mox-ui          if { hdr(host) -i mx.foo } OR { hdr(host) -i mail.foo }
  use_backend mox-autoconf    if { hdr(host) -i mta-sts.foo } OR { hdr(host) -i autoconfig.foo }
  default_backend barbrady

backend mox-ui
  mode http
  server localhost 127.0.0.1:1080 check

backend mox-autoconf
  mode http
  server localhost 127.0.0.1:1081 check
```

I've tried to sniff http, and can confirm that the headers present.
The only noticeable thing - they are in lower case, like in HTTP/2, i.e. `x-forwarded-for: <ip>`.

luckyycode:
Same here with k3s deployment. I got X-Forwarded-For,X-RealIp and other headers set (over proxy protocol, and I can see the real ips in another k3s service), yet nothing here in Mox.

sigio:
Can confirm, same with regular nginx and the following config:
```
    location /webmail/ {
      proxy_set_header Host $host;
      proxy_set_header X-Real-IP $remote_addr;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_set_header X-Forwarded-Proto $scheme;
      proxy_buffering off;
[...]

PR #381, opened 2025-08-30
Log the correct remote address when listeners are forwarded, and allow Forwarded: true for MTS-STS and autoconfig endpoints

lupine:
Currently, we log `http.Request#RemoteAddr` even when `Forwarded: true` is set. This isn't very useful (since in this configuration, the logged ip will probably just be localhost). Instead, respect the `X-Forwarded-For` header when `Forwarded: true` is set.

Additionally, allow `Forwarded: true` to be set for the autoconfig and mta-sts listeners. I didn't bother with the metrics listener, I think that one's fundamentally less interesting since it's not intended to be exposed to the outside world the way autoconfig is.

No tests - I couldn't see offhand that the logger is tested - but I'm running this on my local mox and it's doing the thing:

```
Aug 30 19:34:16 ur.gs mox[200532]: l=debug m="http request" pkg=http httpaccess= handler=admin method=get url=/mail/admin/ host=ur.gs duration="539.381µs" statuscode=200 proto=http/1.1 remoteaddr=10.0.1.181 tlsinfo=plain useragent="Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/60.5 Safari/605.1.15" referer= size=62251 uncompressedsize=311815 cid=198fc40b091

Aug 30 19:35:03 ur.gs mox[200532]: l=debug m="http request" pkg=http httpaccess= handler=(nomatch) method=get url=/ host=mta-sts.ur.gs duration="34.305µs" statuscode=404 proto=http/1.1 remoteaddr=10.0.1.181 tlsinfo=plain useragent="Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/60.5 Safari/605.1.15" referer= size=19 cid=198fc40b098

Aug 30 19:35:35 ur.gs mox[200532]: l=debug m="http request" pkg=http httpaccess= handler=(nomatch) method=get url=/favicon.ico host=autoconfig.ur.gs duration="81.263µs" statuscode=404 proto=http/1.1 remoteaddr=10.0.1.181 tlsinfo=plain useragent="Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/60.5 Safari/605.1.15" referer=https://autoconfig.ur.gs/ size=19 cid=198fc40b09c

```

Closes https://github.com/mjl-/mox/issues/369

### DIFF
--- a/http/web.go
+++ b/http/web.go
@@ -821,7 +821,7 @@ func portServes(name string, l config.Listener) map[int]*serve {
 	}
 	if l.AutoconfigHTTPS.Enabled {
 		port := config.Port(l.AutoconfigHTTPS.Port, 443)
-		srv := ensureServe(!l.AutoconfigHTTPS.NonTLS, false, false, port, "autoconfig-https", false)
+		srv := ensureServe(!l.AutoconfigHTTPS.NonTLS, l.AutoconfigHTTPS.Enabled, false, port, "autoconfig-https", false)
 		if l.AutoconfigHTTPS.NonTLS {
 			ensureACMEHTTP01(srv)
 		}
@@ -851,7 +851,7 @@ func portServes(name string, l config.Listener) map[int]*serve {
 	}
 	if l.MTASTSHTTPS.Enabled {
 		port := config.Port(l.MTASTSHTTPS.Port, 443)
-		srv := ensureServe(!l.MTASTSHTTPS.NonTLS, false, false, port, "mtasts-https", false)
+		srv := ensureServe(!l.MTASTSHTTPS.NonTLS, l.MTASTSHTTPS.Forwarded, false, port, "mtasts-https", false)
 		if l.MTASTSHTTPS.NonTLS {
 			ensureACMEHTTP01(srv)
 		}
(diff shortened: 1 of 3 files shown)


http/web.go
```diff
@@ -349,11 +363,12 @@ func (w *loggingWriter) Done() {
 		slog.Duration("duration", time.Since(w.Start)),
 		slog.Int("statuscode", w.StatusCode),
 		slog.String("proto", strings.ToLower(w.R.Proto)),
-		slog.Any("remoteaddr", w.R.RemoteAddr),
+		slog.String("remoteaddr", w.remoteAddr()),
```
lupine, line 366: We could log the request.RemoteAddr too, as `original-address` or something, but it's fundamentally not very interesting. It might have a marginal use if you're running a cluster of frontend proxies talking to the backend though, so I don't mind if you want it.

mjl-:
@lupine Thanks for the PR. I made a few changes. Mostly to still log the connection remote address as "remoteaddr", but now also logging "clientip" for the x-forwarded-for value. In part because for me "remoteaddr" has the connotation of being a tcp-level address. It also includes a port number, where x-forwarded-for does not (normally), and remoteaddr also has the []-wrapping for ipv6 addresses. The "clientip" field is just the ip.

Would be great if you could check if this works as expected.

PR #402, opened 2025-12-12
webadmin: use state variable for move buttons visibility in webserver handlers

monowii:
Fix move button visibility state when adding handlers in webadmin

Previously, clicking 'Move' after adding a new handler would cause the move buttons to go out of sync, inverting individual states instead of toggling all together. Now uses a shared state variable to keep all handlers synchronized.


mjl-:
Thanks @monowii, I merged it in commit a8cd952fba787cf1efbac414e388444a6cd07838, which also adds the change to the corresponding js file.

PR #413, opened 2026-02-06
added a FreeBSD rc file to run as a service

thenktor:
Hi,

I've added a FreeBSD rc script, so mox can be run as a service.

I'm using similar rc scripts for my own daemons written in Go. The mox directory can be set with `sysrc mox_workdir=/path/to/dir` if it differs from `/home/mox`.


thenktor:
Part of #414

emyfops (approved):
Works perfectly with [mox-mail-server-0.0.15_6](https://cgit.freebsd.org/ports/tree/mail/mox).

quickstart.go
```diff
@@ -1078,7 +1081,7 @@ You can now start the mox container.
 	}
 	fmt.Printf(`
 File ownership and permissions are automatically set correctly by mox when
-starting up. On linux, you may want to enable mox as a systemd service.
+starting up. On Linux or FreeBSD, you may want to enable mox as a service.
```
mjl-:
```suggestion
starting up. You may want to enable mox as a service.
```

mjl-: previously, the sentence mentioned linux mostly for systemd. but it's probably better to just say "start as service", and user will see the systemd file for linux, and the rc script for freebsd.

quickstart.go
```diff
@@ -1098,6 +1101,19 @@ starting up. On linux, you may want to enable mox as a systemd service.
 	sudo systemctl enable $PWD/mox.service
 	sudo systemctl start mox.service
 	sudo journalctl -f -u mox.service # See logs
+`)
+	} else if runtime.GOOS == "freebsd" {
+		xwritefile("mox.freebsd.rc", []byte(moxFreebsdRc), 0555)
```
mjl-: wouldn't 0755 be more common for files in /usr/local/etc/rc.d?

quickstart.go
```diff
@@ -1098,6 +1101,19 @@ starting up. On linux, you may want to enable mox as a systemd service.
...
+		cleanupPaths = append(cleanupPaths, "mox.freebsd.rc")
+		fmt.Printf(`See mox.freebsd.rc for a FreeBSD rc file. To enable and start as root:
```
mjl-:
```suggestion
		fmt.Printf(`See mox.freebsd.rc for a FreeBSD rc file. To enable and start, run as root:
```

mjl-: to make it more clear it's about running the commands as root. otherwise users may think mox runs entirely as root...

quickstart.go
mjl-:
```suggestion
starting up. You may want to enable mox as a service.
```

mjl-: previously, it was about linux with systemd, it now probably makes more sense to say something generic. for linux and now freebsd, the files and instructions make it clear enough how to do it.

quickstart.go
mjl-, line 1107: why 0555? doesn't 0755 make more sense in /usr/local/etc/rc.d?

quickstart.go
mjl-:
```suggestion
		fmt.Printf(`See mox.freebsd.rc for a FreeBSD rc file. To enable and start, run as root:
```

mjl-: to indicate the commands need to be run as root. otherwise admins may start to think all of mox runs as root.

mjl- (changes requested):
great, thanks for the PR! i added a question and some small suggestions.
sorry for the late response.

quickstart.go
thenktor: Does not really matter I think. I've seen 0755 and 0555 permissions there.

mjl-:
excellent, merged in rebased commit 6cbddca56e3ad6c6f471a3e77ee110113f1716ae, thanks!

PR #415, opened 2026-02-14
arc: add ARC RFC 8617 for authenticated received chain protocol

kardianos:
I thought I needed ARC. I was wrong. But I had already created this.
If this is useful, here you go.


mjl-:
Thanks for the PR. Not sure we'll need it. It looks like ARC may be moved to historic, see https://www.ietf.org/archive/id/draft-ietf-dmarc-arc-to-historic-00.html. Perhaps some of the code can be reused for DKIM2.

kardianos:
> Forwarding-driven breakage still dominant: Because email forwarding automatically changes the apparent sender infrastructure (for example, the forwarding system’s IP rather than the original domain), many well-authenticated messages fail DMARC at the final recipient purely due to forwarders. 

This so many times over. All of MS's own spam protection services break the chain and we get reports from them. But is isn't us, it is them. So frustrating. Let's hope DKIM2 has better solutions.

PR #417, opened 2026-02-16
detect which parts need UTF8 and if possible downgrade message

kardianos:
Also fixes incorrect encoding, required for proper downgrade.
Do not forward ext use to server as, but check if SMTPUTF8 is required.


kardianos:
This also fixes #398

kardianos:
@mjl- This PR fixes actual delivery bug I've encountered.

PR #422, opened 2026-04-14
Prefers https over http

TheNotary:
See #423


TheNotary:
Closing till return, too minor to clog up PR list.

PR #425, opened 2026-04-15
Enable healthz endpoint

TheNotary:


http/web.go
```diff
@@ -618,14 +636,18 @@ func portServes(name string, l config.Listener) map[int]*serve {
...
-			s = &serve{nil, nil, tlsNextProtoMap{}, false, false, false, nil, false, nil}
+			s = &serve{nil, nil, tlsNextProtoMap{}, false, false, false, false, nil, false, nil}
 			portServe[port] = s
 		}
 		s.Kinds = append(s.Kinds, kind)
 		if favicon && !s.Favicon {
 			s.ServiceHandle("favicon", accountHostMatch, "/favicon.ico", mox.SafeHeaders(http.HandlerFunc(faviconHandle)))
 			s.Favicon = true
 		}
+		if !s.Healthz {
```
mjl-: what about we only register it on the MetricsHTTP server? seems like the most plausible place to do it. and means we don't expose this info on public endpoints.
would help to update the MetricsHTTP sconf documentation go doc tag to mention that /healthz is available.
i'm more in favor of /health btw. more obvious, and less kubernetes-specific.

docker-compose-integration.yml
```diff
@@ -43,7 +43,7 @@ services:
       - ./testdata/integration/resolv.conf:/etc/resolv.conf:z
       - ./testdata/integration:/integration:z
     healthcheck:
-      test: netstat -nlt | grep ':25 '
+      test: wget -q -O- http://127.0.0.1:80/healthz || exit 1
```
mjl-: is the " || exit 1" needed? seems like a non-successful wget exit status would already be non-successful?

mjl- (changes requested):
thanks for the pr! would be good to get a health endpoint merged. i _think_ it could be better to only expose a health check on the metrics endpoint, see suggestion. what do you think?

http/web.go
TheNotary: I'm on board and thought it was a strange convention too.  The only drawback is minor; users just need to map to the `/metrics/health` endpoint explicitly.

docker-compose-integration.yml
TheNotary: 👍 agreed

TheNotary:
> thanks for the pr! would be good to get a health endpoint merged. i think it could be better to only expose a health check on the metrics endpoint, see suggestion. what do you think?

I think it will work fine that way.  I was split on how to handle it really, it felt heavy to me adding a separate root endpoint so figured follow convention until I had better direction.  

It's unrelated, but as a fun sidenote, the information revealed through `/healthz` can usually be revealed through hitting the other endpoints, it's just less automated.

I'll have some time this weekend to apply those corrections.

TheNotary:
Update commits pushed.

mjl-:
thanks! merged as squashed commit f579d717fbd9b5f7c1ed820ceb6555dfc9ba7bcc.
i made a few small tweaks:
- the health check for localserve in the docker-compose-integration.yml file needed a different ip in the url
- also accept HEAD requests for the health check endpoint. i know some services do HEAD requests for health checks.

PR #428, opened 2026-05-05
Add a "Flagged" button to the webmail refine bar

benjojo:
Some users who have come from gmail environments are big users of the "star" or "flagged" (in IMAP Lingo) feature. I was finding it tricky to easily see what I have flagged (since for me those are normally the biz emails that cannot be answered immediately, so the flag list for me is a (sometimes huge) todo list)

This is my first PR against Mox, I tried to understand the jumbo webmail but it is entirely possible I have missed something here.

I am aware that I can search for "l:\Flagged" in webmail to do this but I think having a quick button is a huge win for people like me coming from gmail/gsuite/google workspace/whatever

I have deployed and tested this change (by applying it on top of v0.0.15) on my own instance and it seems to work fine

### DIFF
--- a/webmail/webmail.js
+++ b/webmail/webmail.js
@@ -2186,6 +2186,10 @@ const refineFilters = (f, notf) => {
 		else if (refine === 'attachments') {
 			f.Attachments = 'any';
 		}
+		else if (refine === 'flagged') {
+			f.Labels = [...(f.Labels || [])];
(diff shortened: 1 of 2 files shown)


mjl-:
Thanks for the report & PR benjojo! I rebased as commit 53d60fad622cb3e6d736bbcb6857148b1d978315, along with a small whitespace tweak.

PR #430, opened 2026-05-09
Restrict TLS cipher suites to GCM/CHACHA20 and bump Go to 1.26.3 for SHA-1 removal

### LINKED ISSUE #409: Setting stricter TLS cipher suite
The NCSC-NL TLS guidelines got an update: see [version 2025-05](https://www.ncsc.nl/api/media/sites/default/files/Publication_TLS-Security%20guidelines-2025-05_ENG.pdf) \*. Currently one can test the configured at https://dev4.internet.nl. I could not find the documentation for Mox to configure TLS cipher suite, e.g. to remove ciphers with CBC.

I did see: https://github.com/mjl-/mox/blob/9795c4e77c15cdabaf97c6f788e96b4ef1013b6c/doc.go#L1559-L1560
but it's only for `mox smtp dial`.

\* Note the current spec says RSA-PKCS1-v1_5 is insufficient, the NCSC-NL is fixing the wording on that point: RSA-PSS is a must in TLS v1.3, but not in TLS v1.2.

pzeinlinger:
Fixes #409.

## Summary

Internet.nl's mail-server check flagged my mox instance for two TLS issues that boil down to Go's default TLS 1.2 configuration:

1. **Insufficient cipher suites**: Go's default cipher list still includes `TLS_ECDHE_ECDSA_WITH_AES_{128,256}_CBC_SHA`, which NCSC-NL rates as insufficient (golang/go#13385 is open and unplanned).
2. **SHA-1 in key exchange**: Go's TLS 1.2 server advertised SHA-1 as a supported signature algorithm — fixed in Go 1.25 per RFC 9155 (golang/go#72883), can be re-enabled with `GODEBUG=tlssha1=1`.

### Changes

- **`mox-/config.go`**: Explicitly set `CipherSuites` on `Config`, `ConfigFallback`, and `ACMEConfig` (right where `MinVersion` is already set) to only allow GCM and CHACHA20-POLY1305 suites that NCSC-NL rates as Sufficient. TLS 1.3 cipher suites remain non-configurable in Go and are always secure.
- **`go.mod`**: Bump from `1.24.0` to `1.26.3` so the SHA-1 signature algorithm change from Go 1.25 takes effect.
- **`docker-compose-integration.yml`** *(separate commit)*: Pebble moved off Docker Hub, so the image reference is updated to `ghcr.io/letsencrypt/pebble`. Pinned to `v2.6.0` — newer pebble releases contain breaking changes for our integration setup. The v2.6.0 image is distroless (no shell, no netstat), so pebble args are passed directly instead of via `sh -c`, the netstat-based healthcheck is dropped, and dependents now wait via `service_started`.

### Allowed cipher suites after this change

```
TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
```

## Test plan

- [x] `go build ./...` succeeds with Go 1.26.3
- [x] Cross-compiled `GOOS=linux GOARCH=arm64 go build` and ran `./mox version` on Linux/aarch64
- [x] Re-ran https://internet.nl/mail/ against an instance built from this branch — 100% score, both "Cipher suites" and "Hash function for key exchange" subtests pass
- [x] `make test-integration` passes against the updated pebble image
- [x] Verified with a real test mail and a Thunderbird IMAP/SMTP connection — no client regressions

## Notes

Happy to split the pebble change into a separate PR if preferred. Also happy to make the cipher suite list configurable via the `TLS` config struct rather than hardcoded, if you'd like to give operators an escape hatch.

### DIFF
--- a/docker-compose-integration.yml
+++ b/docker-compose-integration.yml
@@ -51,7 +51,7 @@ services:
       dns:
         condition: service_healthy
       acmepebble:
-        condition: service_healthy
+        condition: service_started
     networks:
       mailnet1:
         ipv4_address: 172.28.1.10
@@ -76,7 +76,7 @@ services:
       dns:
         condition: service_healthy
       acmepebble:
-        condition: service_healthy
+        condition: service_started
       # moxacmepebble creates tmp-pebble-ca.pem, needed by moxmail2 to trust the certificates offered by moxacmepebble.
       moxacmepebble:
         condition: service_healthy
@@ -104,7 +104,7 @@ services:
       dns:
         condition: service_healthy
       acmepebble:
-        condition: service_healthy
+        condition: service_started
     networks:
       mailnet1:
         ipv4_address: 172.28.1.80
@@ -180,19 +180,16 @@ services:
   # certificate in moxacmepebble and moxmail2.
   acmepebble:
     hostname: acmepebble.example
-    image: docker.io/letsencrypt/pebble:v2.3.1@sha256:fc5a537bf8fbc7cc63aa24ec3142283aa9b6ba54529f86eb8ff31fbde7c5b258
+    image: ghcr.io/letsencrypt/pebble:v2.6.0@sha256:18f6e3326ef8b307ce3e97a8c6f00613e318178450a684415e053507ed1f740f
     volumes:
       - ./testdata/integration/resolv.conf:/etc/resolv.conf:z
       - ./testdata/integration:/integration:z
-    command: ["sh", "-c", "set -ex; mount; ls -l /etc/resolv.conf; chmod o+r /etc/resolv.conf; pebble -config /integration/pebble-config.json"]
+    command: ["-config", "/integration/pebble-config.json"]
     ports:
       - 14000:14000  # ACME port
       - 15000:15000  # Management port
-    healthcheck:
-      test: netstat -nlt | grep ':14000 '
-      interval: 1s
-      timeout: 1s
-      retries: 10
+    # The distroless pebble image has no shell or netstat, so we can't run a
+    # netstat-based healthcheck. Dependents use service_started instead.
     depends_on:
       dns:
         condition: service_healthy
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/mjl-/mox
 
-go 1.24.0
+go 1.26.3
 
 require (
 	github.com/mjl-/adns v0.0.0-20250321173553-ab04b05bdfea
--- a/mox-/config.go
+++ b/mox-/config.go
@@ -837,14 +837,29 @@ func PrepareStaticConfig(ctx context.Context, log mlog.Log, configFile string, c
 				}
 				minVersion = v
 			}
+			// Only allow GCM and CHACHA20-POLY1305 cipher suites for TLS 1.2, excluding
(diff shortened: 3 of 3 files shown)


mox-/config.go
```diff
@@ -837,14 +837,29 @@ func PrepareStaticConfig(ctx context.Context, log mlog.Log, configFile string, c
 				}
 				minVersion = v
 			}
+			// Only allow GCM and CHACHA20-POLY1305 cipher suites for TLS 1.2, excluding
```
mjl-, line 840: i'm wondering if this would cause any interoperability issues leading mail delivery failures. where we aren't accepting a message from an smtp client that only speaks a cbc algorithm.

what about we restrict the ciphersuites only when MinVersion equals TLS1.2, but not when it is older? the default tls minversion is tls1.2, so by default mox admins will be secure. but if there are interoperability issues, they could allow older tls versions again. not recommended of course, but an escape hatch.

also because it seems that the secure algorithms (gcm & chacha) are only available in tls1.2 (https://github.com/golang/go/blob/master/src/crypto/tls/cipher_suites.go, func CipherSuites).

this behaviour should then be documented in config/config.go, in the sconf-doc go struct tag of TLS.MinVersion.

and i think it makes sense to request the list of ciphersuites from crypto/tls.CipherSuites(), and remove the ones we don't want. so we stay more in sync with the go crypto library choices.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/mjl-/mox
 
-go 1.24.0
+go 1.26.3
```
mjl-, line 3: this is interesting. normally, we want to support the supported go toolchains (the current and previous "minor" version).
but indeed sometimes behaviour is enabled based on the go version set here.
in the mean time, the go.mod has version go 1.25, so we are good now.
i think the policy should be to still specify the "previous" go toolchain release, except when we need the latest go toolchain release for a security fix. will write it up.
btw, i don't want to update the toolchain here for every new go toolchain. that could lead to expectations of a new mox release for each new go toolchain release with a fix users should get. we'll assume users build with a recent toolchain.

docker-compose-integration.yml
```diff
@@ -51,7 +51,7 @@ services:
       dns:
         condition: service_healthy
       acmepebble:
-        condition: service_healthy
+        condition: service_started
```
mjl-, line 54: i think a fix for the new pebble is already in place, so i suspect these changes can go.

mjl- (changes requested):
thanks for the pr (sorry for the late reply). we should indeed restrict the cipher suites. i think it would be good to have an escape hatch in case some people run into deliverability issues, see suggestion.

btw, it could also be good to track tls connection state properties in prometheus metrics, at least for the smtp server and smtp client (and probably imap, not necessarily https). that can be considered separately, but if you're interested in implementing that, it would be welcome. (: 

btw, there's now another issue raised by internet.nl's checker, about rsa2048 keys. we should update that too (at least for new installs; updating current/old installs can come later).

PR #443, opened 2026-05-26
Fix integration tests

hilli:
Having problems running `make test-integration`, I found that Pebble changed somewhat since last updates here. 

- The image is no longer on Docker Hub (which prompted me to fix this).
- It is now distoless, so no healthcheck with `netstat`
- There is a `-dnsserver` option, replacing the `resolve.conf` hack, so the `resolve.conf` is not useful for this image (Still used by the other images) and thus not mounted in any longer.
- curl retrieval of the certs have added retries, to compensate for the missing health check

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,8 @@ FROM alpine:latest
 WORKDIR /mox
 COPY --from=build /build/mox /bin/mox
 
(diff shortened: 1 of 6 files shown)


mjl-:
Thanks for the report & fix hilli! Rebased into main branch in most recent commits.

Issue #435, opened 2026-05-14
Webhook contents for "delayed" events don't contain the Error reason


benjojo:
While watching a email sent that was sent to a MX that was down, I noticed that while it is trying the "error" is `delivering...` : 

<img width="1286" height="316" alt="Image" src="https://github.com/user-attachments/assets/e3c5ff51-a46c-4665-ac98-5eea898b7fab" />

And when the timeout happens, it turns into a "real" error"

<img width="1286" height="307" alt="Image" src="https://github.com/user-attachments/assets/2bfc83f9-e6d4-4aa5-8604-605b70ffa0f6" />

but the `delivering...` error is the one that is delivered over the webhook, as seen in my tracking table:

```
Name             |Value                                           |
-----------------+------------------------------------------------+
message_id       |<b48ebbaf-3897-4d5d-aee6-cc7060fd91a4@bgp.tools>|
notification_uuid|                                                |
time_of_feedback |2026-05-14 15:17:26.000                         |
eventType        |delayed                                         |
DSN              |0                                               |
QueueMsgID       |222                                             |
SMTPCode         |0                                               |
SMTPEnhancedCode |                                                |
Error            |delivering...                                   |
EmailSubject     |[bgp.tools] Edit Request Rejected               |
```

benjojo:
This was fixed on 0.16

PR #444, opened 2026-05-29
Build error webhook payloads with up to date status information

benjojo:
When mox attempts an outbound delivery the prepare func adds a MsgResult for the delivering state to the results array, so that there is something in the DB for when the deliery begins

When a (in this case tempoary) failure happens though, it uses the wrong array with stale data about what happened to build the webhook delivery, the umsgs array /should/ (and seems to when tested) have the real error in it.

This stops status webhooks from being delivered with just "Delivering..." in the error string.

tag: https://github.com/mjl-/mox/issues/435

### DIFF
--- a/queue/dsn.go
+++ b/queue/dsn.go
@@ -191,10 +191,10 @@ func failMsgsTx(qlog mlog.Log, tx *bstore.Tx, msgs []*Msg, dialedIPs map[string]
 			return nil
 		}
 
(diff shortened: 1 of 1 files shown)


mjl-:
clear, thanks! merged in commit 18ecfcdf7a23504c92ef604fc7e599f13ca9ba44

PR #451, opened 2026-06-29
Document Account method naming scheme for broadcast responsibility

Yjlion:
Replace the TODO comment with clear documentation explaining the consistent pattern used throughout the Account type:
- Methods accepting *bstore.Tx are low-level operations where the caller must broadcast changes after committing the transaction
- Methods without *bstore.Tx are high-level operations that handle broadcasting internally

This makes it explicit to maintainers and API users which methods require external change broadcasting and which handle it themselves.

Claude-Session: https://claude.ai/code/session_01BpmSQS4VpPG95qrCdhFkNF

### DIFF
--- a/store/account.go
+++ b/store/account.go
@@ -19,7 +19,21 @@ database.
 */
 package store
 
-// todo: make up a function naming scheme that indicates whether caller should broadcast changes.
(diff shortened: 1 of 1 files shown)


Yjlion:
error

PR #454, opened 2026-07-13
Introduce introbox folder for senders with no reputation

asciimoo:
Adds an optional Introbox for accepted messages from senders with no established reputation.

Messages remain neutral in Introbox. Moving or copying them to their intended mailbox records positive reputation, while moving them to Junk records negative reputation.
Future messages are routed using that reputation.

Includes account configuration, webmail, IMAP and web API support, validation, documentation, and regression tests.

It is possible to make introbox a subfolder on the web UI by naming it `Inbox/Introbox`.


mjl-:
Thanks, I had a first look, and it looks pretty good!
I need some more time to convince myself that the condition for delivering to the Introbox is correct (doesn't cause unwanted side effects).

I'm not sure the separate JunkFlagsForMailboxMove is needed. If you "handle" a message by moving it to archive or junk, it will already get its nonjunk/junk flags. In the webmail you automatically set the nonjunk flag if you read a message for 5 seconds.

In Account.IntroboxSave, we should probably first do the save of the account. That will verify the mailbox is correct. Then creating the mailbox doesn't need to verify again.

We should probably check that Introbox isn't Inbox. There's no point in doing that, and it may cause confusion (or highlight confusion on the users part).

asciimoo:
Thanks for the review.

> I need some more time to convince myself that the condition for delivering to the Introbox is correct (doesn't cause unwanted side effects).

I'm not entirely sure if it is the right approach, this part of the internals are still a bit unclear to me. I borrowed this idea from the `firstTimeSenderDelay` feature, it uses the same check to identify first time senders. Let me know if there is any better way to identify senders with no reputation.

> I'm not sure the separate JunkFlagsForMailboxMove is needed.

Perhaps I did mess up something, but my observation was that moving a message from the introbox to the inbox removes the non-junk flag from it, that's why I set the flag explicitly.

> In Account.IntroboxSave, we should probably first do the save of the account.

Alright.

> We should probably check that Introbox isn't Inbox. There's no point in doing that, and it may cause confusion (or highlight confusion on the users part).

You are right.

---

I've fixed both save and name check issues.

mjl-:
Thanks for the changes! I had another look and made a diff. I attached it
here (I can't have a fork of my own project in github, and I prefer not to
have WIP-commits accessible in this repo for build tools; as zip since github
doesn't support diff files (!)). Need a better way to suggest code changes...

[introbox.diff.zip](https://github.com/user-attachments/files/30152100/introbox.diff.zip)

Comments:
- Removed the introbox feature from readme to not make that list too long.
  The explanation on the website should be enough.
- I added "introbox" (the expected name for these mailboxes) to the
  "automatic junk flags" neutral mailbox regexp. So that we don't need a
  special case in store.Account.MessageAdd.
- For incoming deliveries over smtp, I think we can't use the same check as
  for "first time sender delay". That delay doesn't work on unique address.
  It doesn't apply to mail servers that have sent before because they have a
  reputation based on dkim/spf/ip-address. E.g. gmail.com. We don't want to
  repeatedly slow down gmail mail server, or they won't like us. The
  introbox should apply to unique addresses that are sending. We don't want
  any sender from gmail.com to make it straight to the inbox, a lot of spam is
  coming through the freemail providers (outlook.com as well). I changed the
  condition to apply the Introbox if we don't have a unique email address with
  reputation, either a previous incoming message, or a message we previously
  sent to.
- The change in webmail is to make it clearer that "introbox" isn't one of
  the well-known special-use flags used in imap.

I think we should also consider special handling for message threads. If I get
a first message from personA@example.com, that has a CC to
personB@example.com (or perhaps it's To personB@ and I'm in the CC), and I
move that message to the Inbox (or I already know personA) and then get a
reply from personB@example.com as first correspondence, then typically the
message is allowed in due to a known domain (and likely also due to a known
dkim/spf reputation of the sending server), but it would be delivered to the
Introbox. As a user, I probably would like to see it delivered to my regular
inbox (I may only check my introbox once a day or once a week). So perhaps we
could do something like this: For message that would go to the introbox, look
at the parent message of its thread (if any), and if that is in the (same?)
original destination mailbox or is nonjunk, then don't deliver to introbox
but deliver as usual.
Currently, we do thread assigning in store.Account.MessageAdd, so after the
decision has been made where to deliver to. Perhaps we can make (a simplified
version of) the thread assigning exported and use it from smtpserver. Or pass
in an indication to MessageAdd to handle the mailbox-assigning.
If we do it early, in smtpserver, we could also do this for regular
incoming mail (allow it in if it's coming from an address that is a CC/BCC of
the parent message (or an ancestor message). Let me know if you want me to
give it a try to make this change.

Btw, the JunkFlagsForMailboxMove looks good.

Was moving CheckMailboxName to the config/ package needed? We should probably
change all calls from store.CheckMailboxName to config.CheckMailboxName?

I didn't update the tests btw, I likely made them fail now...

mjl-:
I had another thought that I forgot to mention: This should probably only work for messages delivered to the inbox. Users can have rulesets that deliver some messages to other mailboxes. E.g. if you have multiple addresses in your account, and arrange for some messages to be delivered to a different mailbox. It would probably be confusing to some of those messages delivered to a single introbox.

In the future, we could extend this functionality to specify an introbox per mailbox that a ruleset would match. Perhaps a new field "Introbox" below "Mailbox" in the ruleset, below https://dev.xmox.nl/config/#cfg-domains-conf-Accounts-x-Destinations-x-Rulesets-dash-Mailbox.

asciimoo:
> I can't have a fork of my own project in github, and I prefer not to have WIP-commits accessible in this repo for build tools;

A possible solution is creating review branches with a specific name format (e.g. `review-454` / `review-introbox`) and disabling the unwanted build tools for the matching branch name pattern.

> Was moving CheckMailboxName to the config/ package needed?

The mox package cannot import store, because store already imports mox.

> We should probably change all calls from store.CheckMailboxName to config.CheckMailboxName?

I agree.

---

Thank you for the thorough review. I've applied your patch, fixed the tests and replaced the `store.CheckMailboxName` calls with the new `config.CheckMailboxName`.

How do you want to move forward with the suggested improvements about thread checks and thread assignment? Should I implement it in this PR?

mjl-:
> > I can't have a fork of my own project in github, and I prefer not to have WIP-commits accessible in this repo for build tools;
> 
> A possible solution is creating review branches with a specific name format (e.g. `review-454` / `review-introbox`) and disabling the unwanted build tools for the matching branch name pattern.

it's the go module proxy i expect will be able to fetch commits from the
branches. should test it. anyway, i can just make a new repo and push the
whole git history there.

> Thank you for the thorough review. I've applied your patch, fixed the tests and replaced the `store.CheckMailboxName` calls with the new `config.CheckMailboxName`.

great!

> How do you want to move forward with the suggested improvements about thread checks and thread assignment? Should I implement it in this PR?

you could give it a shot, but it could become a bit complicated. we could also
just start using the introbox functionality as you've implemented it now and
consider it finished, and get some experience with using it, and adding the
thread matching in the future (if we ever need to).

asciimoo:
> you could give it a shot, but it could become a bit complicated. we could also
just start using the introbox functionality as you've implemented it now and
consider it finished, and get some experience with using it, and adding the
thread matching in the future (if we ever need to).

I'd prefer not adding more complexity with this PR if you consider the feature complete without the thread related changes.

mjl-:
hi @asciimoo! i merged the MR with a few more changes in squashed commit https://github.com/mjl-/mox/commit/9928c0aaf488aada87e17b1327621acc388b2c1c, with a followup commit https://github.com/mjl-/mox/commit/2b548027a704c31c1aed326caf3e23c3f141d056. i'm running it myself now to see how it goes. i'll send you an email!

asciimoo:
Great, thank you!

PR #455, opened 2026-07-27
webmail: don't intercept cmd+key browser shortcuts

ajaspers:
Previously, the message-list key handler intercepted move keys (l for page down, etc.) and called preventDefault whenever alt wasn't held. On macOS, cmd+l (jump to address bar) gets swallowed, along with several others like cmd+u and cmd+p.

cmd/meta isn't used for any move action, so skip the move-key handler entirely when it's held and let the browser handle those keys.


mjl-:
Thanks @ajaspers, merged this PR after rebase with commit 87a14f7ba6f5f5c9c58e34a517314f102c368018.

PR #457, opened 2026-08-02
imapserver: fix command-state check precedence

ajaspers:
The state check read `ok3 && c.state == stateAuthenticated || c.state == stateSelected`, which parses as `(ok3 && authenticated) || selected`, making stateSelected a catch-all that allowed commands meant only for other states (e.g. starttls/authenticate/login). Parenthesize the condition so authenticated-state commands are gated by the connection state.

This also surfaced that CHECK was registered in the commands map but missing from commandsStateSelected. Add it so CHECK keeps working with this change.


mjl-:
Thanks @ajaspers, great find. This allows violated all kinds of expectations. I merged through rebased commit 1d1a358b4102c74016bc3287462dfd28095ed6ba.

PR #441, opened 2026-05-22
feat: reject mail when From address is invalid

acheong08:
Fixes a spam vector which bypasses SPF checks.
Closes #440

Issue #440, opened 2026-05-22
Reject invalid email addresses


acheong08:
I've been getting a ton of spam lately.

Here's a header:

```eml
X-Mozilla-Status: 0001
X-Mozilla-Status2: 00000000
X-Mox-Reason: no-bad-signals; not using any dmarc result;
	address/dkim/spf/ip-based reputation (inconclusive, no address/spf/dkim/ip
	 reputation);
	content: not junk, spamscore 0.00, threshold 0.25 (stricter due to missing
	 tls) (ham words: Subject:database 0.010, accountproceed 0.010,
	 Subject:server 0.010, Subject:mails 0.010, Subject:email 0.010,
	 Subject:hold 0.010, Subject:have 0.010, Subject:been 0.010, From:domain
	 0.010, From:email 0.010), (spam words: generated 0.808, dear 0.641,
	 reserved 0.634, rights 0.612);
	no known reputation and no bad signals;
Delivered-To: acheong@duti.dev
Return-Path: <>
Authentication-Results: mail.duti.dev;
	iprev=pass (without dnssec) policy.iprev=35.252.109.202;
	dkim=none reason="no dkim signatures";
	spf=none;
	dmarc=none
Received-SPF: none (no domain, ehlo is an ip literal and mailfrom is empty)
	client-ip=35.252.109.202; envelope-from="\"\"@10.88.0.3"; helo=10.88.0.3;
	receiver=mail.duti.dev; identity=""
Received: from
	[10.88.0.3] (202.109.252.35.bc.googleusercontent.com [35.252.109.202]) by
	mail.duti.dev ([2.58.53.24]) via tcp with ESMTP id 0WBNuTOgnAHCFv3JX64NJg
	for <acheong@duti.dev>; 22 May 2026 11:27:16 +0000
Content-Type: multipart/related; boundary="===============3702814871923690882=="
MIME-Version: 1.0
From: "(domain)" <(email)(domain)>
To: acheong@duti.dev
Subject: =?utf-8?q?mails_have_been_put_on_hold_in_the_email_server_database=2E?=
X-Priority: 2
```

The From address is not valid, and therefore SPF and DKIM are bypassed.

acheong08:
<img width="1076" height="1330" alt="Image" src="https://github.com/user-attachments/assets/be361b9d-6dd0-4ae9-a760-1b4b456668bb" />

This is what the spam looks like

mjl-:
I agree that not having a valid single From header is suspicious.

[...]

PR #459, opened 2026-08-08
message: refuse to parse multiple From headers

ajaspers:
An RFC5322 message has exactly one From header (../rfc/5322:1145). Multiple From headers are only allowed by the obsolete syntax, which explicitly leaves their interpretation unspecified (../rfc/5322:1993).

The check in From() counted addresses within a single header, so a message with two From headers passed. DKIM signs the last instance of a repeated header (../rfc/6376:2320). So a sender could get dmarc=pass with header.from of a domain they control, on a message with an additional unvalidated From header.

### DIFF
--- a/message/from.go
+++ b/message/from.go
@@ -1,6 +1,7 @@
 package message
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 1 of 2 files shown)


mjl-:
thanks, great find again! i also added commit d407e17b054e716a1b61573a6c480d8e309bf0c4 which should ensure the multiple headers for "from"/"to"/"cc" etc are used when parsing the envelope.
i'm curious how you're finding these issues!

ajaspers:
Thanks @mjl-! I found this and #457 via LLM code analysis. I've been (slowly) reviewing the LLM findings and creating PRs for real issues. #455 was found via normal usage of Mox webmail on macOS.

If you want to take a look at the rest of the LLM findings that I haven't had a chance to go through yet, I can send them over. I don't think any of them are critical issues.

Somewhat related to this PR: I'm curious what you think about #440 / #441 since lacking a From header is an extremely strong spam signal.

mjl-:
> Thanks @mjl-! I found this and #457 via LLM code analysis. I've been (slowly) reviewing the LLM findings and creating PRs for real issues. #455 was found via normal usage of Mox webmail on macOS.
> If you want to take a look at the rest of the LLM findings that I haven't had a chance to go through yet, I can send them over. I don't think any of them are critical issues.

Nice! I'm curious about what the LLM is finding. I wouldn't mind seeing
findings. Though I probably won't have time to really look into them. So
at least please keep doing what you're doing. (:

I would also be interested in hearing about your process, and the experiences
you're having with testing mox this way. For example, can the LLM easily run
mox to verify the finding is real? Or is there something about mox that makes
that hard to do for LLMs? (And is there a way to make it easier?)

> Somewhat related to this PR: I'm curious what you think about #440 / #441 since lacking a From header is an extremely strong spam signal.

I'll comment there!

PR #463, opened 2026-08-16
scram: fix parser offset in xvalue

ajaspers:
A value terminated by a comma reset the parser offset to the relative offset (p.o = o) instead of advancing it (p.o += o), corrupting the parser state.

This affects both the client-first message (NewServer) and the client-final message (Finish), which each parse trailing extensions.

Add a test which fails before this fix and passes after.


mjl-:
Excellent find & fix, thanks!

PR #466, opened 2026-08-28
feat: inbounds proxy protocol v1/v2 support

### LINKED ISSUE #370: Support PROXY protocol
I think it might be useful to support [PROXY protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt).
I saw several implementations for Go, so it should not be big deal.

qingyiwebt:
Add support for PROXY Protocol, which allows us to use non-standard ports in some cases, and proxy them with HAProxy or OpenResty.

This is useful when the mox server cannot directly listen on the standard mail ports, for example because they are blocked by the ISP, and another server is used as a TCP proxy. It would also be useful for clustered deployments.

Also, closes #370

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -158,29 +158,34 @@ type Listener struct {
 
 		FirstTimeSenderDelay *time.Duration `sconf:"optional" sconf-doc:"Delay before accepting a message from a first-time sender for the destination account. Default: 15s."`
 
-		TLSSessionTicketsDisabled *bool `sconf:"optional" sconf-doc:"Override default setting for enabling TLS session tickets. Disabling session tickets may work around TLS interoperability issues."`
+		TLSSessionTicketsDisabled *bool          `sconf:"optional" sconf-doc:"Override default setting for enabling TLS session tickets. Disabling session tickets may work around TLS interoperability issues."`
+		ProxyProtocol             *ProxyProtocol `sconf:"optional" sconf-doc:"Require a PROXY protocol v1 or v2 header on the dedicated SMTP port. Only connections from TrustedProxies are accepted. HTTPS/ALPN is not affected."`
 
 		DNSBLZones []dns.Domain `sconf:"-"`
 	} `sconf:"optional"`
 	Submission struct {
 		Enabled           bool
-		Port              int  `sconf:"optional" sconf-doc:"Default 587."`
-		NoRequireSTARTTLS bool `sconf:"optional" sconf-doc:"Do not require STARTTLS. Since users must login, this means password may be sent without encryption. Not recommended."`
+		Port              int            `sconf:"optional" sconf-doc:"Default 587."`
+		NoRequireSTARTTLS bool           `sconf:"optional" sconf-doc:"Do not require STARTTLS. Since users must login, this means password may be sent without encryption. Not recommended."`
+		ProxyProtocol     *ProxyProtocol `sconf:"optional" sconf-doc:"Require a PROXY protocol v1 or v2 header on the dedicated Submission port. Only connections from TrustedProxies are accepted. HTTPS/ALPN is not affected."`
 	} `sconf:"optional" sconf-doc:"SMTP for submitting email, e.g. by email applications. Starts out in plain text, can be upgraded to TLS with the STARTTLS command. Prefer using Submissions which is always a TLS connection."`
 	Submissions struct {
 		Enabled        bool
-		Port           int  `sconf:"optional" sconf-doc:"Default 465."`
-		EnabledOnHTTPS bool `sconf:"optional" sconf-doc:"Additionally enable submission on HTTPS port 443 via TLS ALPN. TLS Application Layer Protocol Negotiation allows clients to request a specific protocol from the server as part of the TLS connection setup. When this setting is enabled and a client requests the 'smtp' protocol after TLS, it will be able to talk SMTP to Mox on port 443. This is meant to be useful as a censorship circumvention technique for Delta Chat."`
+		Port           int            `sconf:"optional" sconf-doc:"Default 465."`
+		EnabledOnHTTPS bool           `sconf:"optional" sconf-doc:"Additionally enable submission on HTTPS port 443 via TLS ALPN. TLS Application Layer Protocol Negotiation allows clients to request a specific protocol from the server as part of the TLS connection setup. When this setting is enabled and a client requests the 'smtp' protocol after TLS, it will be able to talk SMTP to Mox on port 443. This is meant to be useful as a censorship circumvention technique for Delta Chat."`
+		ProxyProtocol  *ProxyProtocol `sconf:"optional" sconf-doc:"Require a PROXY protocol v1 or v2 header on the dedicated Submissions port. Only connections from TrustedProxies are accepted. HTTPS/ALPN is not affected."`
 	} `sconf:"optional" sconf-doc:"SMTP over TLS for submitting email, by email applications. Requires a TLS config."`
 	IMAP struct {
 		Enabled           bool
-		Port              int  `sconf:"optional" sconf-doc:"Default 143."`
-		NoRequireSTARTTLS bool `sconf:"optional" sconf-doc:"Enable this only when the connection is otherwise encrypted (e.g. through a VPN)."`
+		Port              int            `sconf:"optional" sconf-doc:"Default 143."`
+		NoRequireSTARTTLS bool           `sconf:"optional" sconf-doc:"Enable this only when the connection is otherwise encrypted (e.g. through a VPN)."`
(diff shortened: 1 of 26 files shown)


config/config.go
```diff
@@ -158,29 +158,34 @@ type Listener struct {
 
 		FirstTimeSenderDelay *time.Duration `sconf:"optional" sconf-doc:"Delay before accepting a message from a first-time sender for the destination account. Default: 15s."`
 
-		TLSSessionTicketsDisabled *bool `sconf:"optional" sconf-doc:"Override default setting for enabling TLS session tickets. Disabling session tickets may work around TLS interoperability issues."`
+		TLSSessionTicketsDisabled *bool          `sconf:"optional" sconf-doc:"Override default setting for enabling TLS session tickets. Disabling session tickets may work around TLS interoperability issues."`
+		ProxyProtocol             *ProxyProtocol `sconf:"optional" sconf-doc:"Require a PROXY protocol v1 or v2 header on the dedicated SMTP port. Only connections from TrustedProxies are accepted. HTTPS/ALPN is not affected."`
```
mjl-: why is https/alpn mentioned here? this is just forwarding the tcp connection, right? tls (alpn) would be done by mox, and https is not involved in smtp.

config/config.go
```diff
@@ -222,6 +227,13 @@ type Listener struct {
 	} `sconf:"optional" sconf-doc:"All configured WebHandlers will serve on an enabled listener. Either ACME must be configured, or for each WebHandler domain a TLS certificate must be configured."`
 }
 
+// ProxyProtocol configures strict PROXY protocol handling for one dedicated
```
mjl-: probably better not to mention smtp/imap. i would suspect it covers all protocols.

imapserver/server.go
```diff
@@ -444,7 +445,7 @@ func listen1(protocol, listenerName, ip string, port int, tlsConfig *tls.Config,
 			}
 
 			metricIMAPConnection.WithLabelValues(protocol).Inc()
-			go serve(listenerName, mox.Cid(), tlsConfig, conn, xtls, noTLSClientAuth, noRequireSTARTTLS, false, "")
+			go serve(listenerName, mox.Cid(), tlsConfig, conn, xtls, noTLSClientAuth, noRequireSTARTTLS, false, "", proxyConfig)
```
mjl-: i think it would be better if we would wrap the tcp listener in a proxylistener just after `mox.Listen` earlier. so that each `Accept` returns a net.Conn that returns the remote net.Addr. then serve() doesn't have to know anything about the proxy protocol. the test code would perhaps be harder, or it's probably enough to do without, it's would just be a matter of a wrapping call for the listener.

i suppose a wrapping listener would have to accept tcp connections, then read the protocol message in a goroutine, and then pass on the initialized connection to a caller to Accept. and a Close on that listener would have to clean up all in-progress handshakes. not great, but the abstraction seems worth the hassle.

with the "wrapping listener" approach, it would be just as easy to also provide proxy support for the http endpoints. and perhaps mox.Listen should even just get a parameter for the proxy config, so we can handle it in just that place.

btw, i saw the wrapping listener from github.com/pires/go-proxyproto just now, and it doesn't seem correct. it doesn't read the proxy header before returning a connection...

imapserver/server.go
```diff
@@ -771,7 +772,26 @@ var cleanClose struct{} // Sentinel value for panic/recover indicating clean clo
 // preauthenticated.
 //
 // The connection is closed before returning.
-func serve(listenerName string, cid int64, tlsConfig *tls.Config, nc net.Conn, xtls, noTLSClientAuth, noRequireSTARTTLS, viaHTTPS bool, preauthAddress string) {
+func serve(listenerName string, cid int64, tlsConfig *tls.Config, nc net.Conn, xtls, noTLSClientAuth, noRequireSTARTTLS, viaHTTPS bool, preauthAddress string, proxyConfigs ...*config.ProxyProtocol) {
```
mjl-: this comment may not be relevant if the listener already handled the proxy protocol, but still:
curious why this change would allow multiple config.ProxyProtocol parameters (and ignore all but the first).
if it's for not having to change other callers: in the mox code base i want to stay explicit and consistent, and we should just update all other callers.

proxyprotocol/conn.go
```diff
@@ -0,0 +1,117 @@
...
+	"bufio"
+	"errors"
+	"fmt"
+	"net"
+	"time"
+
+	proxyproto "github.com/pires/go-proxyproto"
+)
+
+const headerTimeout = 30 * time.Second
```
mjl-: 30s seems a bit long? i would expect the proxy server to be run relatively close to mox, and immediately send the entire header in a single packet right after opening the connection. is there some best practice guide for this value? or how is proxy support deployed normally?
on the other hand, perhaps it's not a problem for this to be a relatively long timeout?

proxyprotocol/conn.go
```diff
@@ -0,0 +1,117 @@
...
+// PROXY header. All other operations, including closing and deadlines, are
+// delegated to the underlying connection.
+type Conn struct {
+	net.Conn
+	reader     *bufio.Reader
+	remoteAddr net.Addr
+	localAddr  net.Addr
+}
+
+// Read first drains bytes already buffered while parsing the PROXY header,
```
mjl-: only handling the first reads from a buffer, and the rest directory from the underlying Conn sounds good. but the implementation does not match comment: all reads go through the bufio.Reader. it seems unnecessary to hold on to the bufio.Reader, with the memory it uses. the smtp & imap servers will have their own buffering again.
in practice, the imap & smtp servers write the first message in their protocols. so the header can usually be read (even if buffered) without having any leftover data that needs to be read. it may be different for http.

we should probably also keep track of the connection close status. if net.Conn is closed, Read will now still return data, not correct.

proxyprotocol/conn.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	remoteAddr, localAddr := conn.RemoteAddr(), conn.LocalAddr()
+	if !header.Command.IsLocal() {
+		remoteAddr, localAddr, _ = header.TCPAddrs()
+	}
+	return &Conn{Conn: conn, reader: reader, remoteAddr: remoteAddr, localAddr: localAddr}, nil
+}
+
+func validateHeader(header *proxyproto.Header) error {
```
mjl-, line 148: should we also reject connections where the remote proxy has already tls? likely a misconfiguration of the proxy server. we want mox to do the tls.
and perhaps the same applies to other TLVs from proxy v2?

proxyprotocol/conn.go
```diff
@@ -0,0 +1,117 @@
...
+		if network != nil && network.Contains(peerIP) {
+			trusted = true
+			break
+		}
+	}
+	if !trusted {
+		return nil, fmt.Errorf("proxy peer %s is not trusted", peerIP)
+	}
+
+	reader := bufio.NewReader(conn)
```
mjl-: i think the proxy protocol docs mention something about the max size of a header. we probably don't need a full bufio.Reader, with its large buffer.

also, we could perhaps have a simple buffered channel where we stash some of these readers for reuse.

mjl- (changes requested):
thanks for the PR @qingyiwebt!
it would be good to get proxy protocol support.
see the comments for requests, suggestions and questions.

btw, i would prefer an implementation of the proxy protocol in mox, with cross-referencing to the protocol specifications like we have for other protocols...

config/config.go
qingyiwebt: Should I keep it for listeners that support `EnabledOnHTTPS`, or remove all wording like this?

proxyprotocol/conn.go
qingyiwebt: For my situation, mail ports are banned from my US server, but they are open on my HK server (for privacy, i can't deploy mox on the HK server which is shared with my friends,  so i just ran openresty there as a proxy.)

The connection between these two servers isn't stable, so I set a 30s timeout. (i had seen that they can't ping each other randomly. now i use vpn like tailscale for them, but the latency might be worse)

config/config.go
mjl-: Ah, good one. I think we remove the wording entirely. In my mind, "proxy support" is just a way to forward a tcp connection that passes on the original remote address. I think it would be good to explain that on the comment/doc for the ProxyProtocol type in the config. That comment/doc could also say that the forwarding proxy should not do TLS, but should only forward the raw tcp connections.

Also, the HTTPS support for submission & imap isn't that commonly used, so the risk of confusing (many) users is probably very low.

proxyprotocol/conn.go
mjl-: clear, thanks, let's keep 30s!

Issue #426, opened 2026-04-17
IMAP APPEND can fail during preview generation for quoted-printable text/plain bodies that decode into one very long line


cukupupas:
I hit:  bufio.Scanner: token too long

  At first this looked like a raw line-length problem, but the sample message suggests otherwise. The text/plain part is
  Content-Transfer-Encoding: quoted-printable, with many short physical lines ending in =. After MIME decoding, those
  soft line breaks disappear, so the preview code ends up scanning one huge logical line and bufio.Scanner fails with
  token too long.

  So this is not necessarily a malformed raw message with oversized source lines. It can happen with a quoted-printable
  body that is physically wrapped correctly but decodes into effectively one long line.

  This seems too strict. Preview generation is non-essential, and I would expect the message to be accepted with an
  empty/unset preview instead of rejecting the whole APPEND.

  A likely reproduction is a text/plain part encoded as quoted-printable where most lines end with soft breaks (=), so
  decoding collapses them into one long logical line.

mjl-:
https://github.com/mjl-/mox/commit/be2b665318823a932fcdc824b919c031d8215bab from just now should fix the issue.

I also tried @hilli's PR #467, and the approach is good, but that seems to have an issue with potentially very long ">"-quoted text before the new text (the end of the quoted text would be returned), I'll add some more details about it in the PR.

PR #467, opened 2026-08-30
message: don't fail the preview on a line longer than the scanner's buffer

hilli:
For issue #426 - I ran into the same issue syncing my mails into mox.

previewText scans with a default bufio.Scanner, so a logical line over 64kb makes Scan stop and Err return ErrTooLong. previewText returned that, and Part.Preview turns any error other than moxio.ErrLimit into an error for the TEXT/PLAIN case, so the message was rejected: delivery failed, and IMAP APPEND answered "NO ... generating preview: making preview from text part: bufio.Scanner: token too long".

Such messages are not malformed. Quoted-printable soft line breaks mean a body wrapped at 78 columns on the wire decodes into a single logical line of any length, and format=flowed does the same. One example that triggers this is wrapped correctly at 78 columns and decodes to a single line of 65572 bytes, 36 over the limit. It also carries a text/html alternative of under 6kb, which would have been forgiven, since the TEXT/HTML case logs and continues where TEXT/PLAIN aborts.

Refusing a message over this seems out of proportion: a preview is advisory, and RFC 8970 does not require that one exists.

So scan with a split function that hands back a full buffer as a line instead of giving up. Nothing is lost by truncating: only the first lines are examined and at most 256 characters are returned, so the bytes past the buffer could never have reached the result. Messages that fit today are unaffected, since the split only differs from bufio.ScanLines once the buffer is full.

### DIFF
--- a/message/preview.go
+++ b/message/preview.go
@@ -93,6 +93,7 @@ func previewText(r io.Reader) (string, error) {
 	// We look quite a bit of lines ahead for trailing signatures with trailing empty lines.
 	var lines []string
 	scanner := bufio.NewScanner(r)
+	scanner.Split(scanPreviewLines)
 	ensureLines := func() {
 		for len(lines) < 10 && scanner.Scan() {
 			lines = append(lines, strings.TrimSpace(scanner.Text()))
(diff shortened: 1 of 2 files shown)


mjl-:
Thanks for the PR @hilli!

The approach looked good, but the end of ">"-quoted text (past the first max scan buffer size) could be returned as preview text. Fixing that by ignoring remaing text on a line in the split function felt a bit tricky (I tried). So I went with the simpler approach of just using a bufio.Reader with ReadString('\n'), and rely on the limitreader to have at least some bound of allocations... See https://github.com/mjl-/mox/commit/be2b665318823a932fcdc824b919c031d8215bab.

hilli:
Perfect, thanks :)

PR #470, opened 2026-09-05
dnsbl monitor: don't set mox_dnsbl_ips_success to 0 on temporary DNS errors

### LINKED ISSUE #445: mox_dnsbl_ips_success goes to 0 when DNS server errors
Noticed this as a surprise Sunday morning alert:

```
l=debug m="dns lookup result" err="lookup 47.223.230.185.sbl.spamhaus.org. on [xxxx::]:53: server misbehaving" pkg=dnsblmonitor type=ip network=ip4 host=47.223.230.185.sbl.spamhaus.org. resp=[] authentic=false duration=142.569485ms

l=debug m="dnsbl lookup result" err="dnsbl: dns error: lookup 47.223.230.185.sbl.spamhaus.org. on [xxxx::]:53: server misbehaving" pkg=serve pkg=dnsbl zone=sbl.spamhaus.org ip=185.230.223.47 status=temperror explanation= duration=142.637667ms

l=error m="dnsbl monitor lookup" err="dnsbl: dns error: lookup 47.223.230.185.sbl.spamhaus.org. on [xxxx::]:53: server misbehaving" pkg=serve ip=185.230.223.47 zone=sbl.spamhaus.org expl= status=temperror
```

resulted in

```
# HELP mox_dnsbl_ips_success DNSBL lookups to configured DNSBLs of our IPs.
# TYPE mox_dnsbl_ips_success gauge
mox_dnsbl_ips_success{ip="185.230.223.47",zone="bl.spamcop.net"} 1
mox_dnsbl_ips_success{ip="185.230.223.47",zone="sbl.spamhaus.org"} 0
```

This seems like the wrong behavior in this situation

Huang-404-Q:
# Don't set `mox_dnsbl_ips_success` to 0 on temporary DNS errors

## What this fixes

The `mox_dnsbl_ips_success` gauge drops to `0` when a DNSBL lookup hits a
temporary DNS error — e.g. a resolver answering "server misbehaving":

```
l=error m="dnsbl monitor lookup" err="dnsbl: dns error: lookup 47.223.230.185.sbl.spamhaus.org. ... server misbehaving" ... status=temperror
```

resulting in:

```
mox_dnsbl_ips_success{ip="185.230.223.47",zone="sbl.spamhaus.org"} 0
```

A temperror says nothing about whether the IP is in the block list, so this
is a false alarm: the issue reporter got a surprise Sunday-morning alert
from it (#445).

## Why it happens

The DNSBL monitor only ever sets the gauge to `1` on `dnsbl.StatusPass`;
every other status — including `dnsbl.StatusTemperr` — falls through to
`Set(0)`. `StatusTemperr` is its own outcome: any A-lookup error other than
NXDOMAIN (timeout, SERVFAIL, "server misbehaving") maps to it, distinct
from `StatusFail`.

## The fix

Skip the gauge update on `StatusTemperr`, so the metric keeps the last
known value until a conclusive result arrives. `pass` → `1` and `fail` →
`0` semantics are unchanged, and the separate `-1` "broken" value (set
when listing our own IPs fails) is untouched.

One side effect worth noting: a zone/ip combination that has only ever
seen temperrors has no gauge series at all, instead of one showing `0`.
For alerting on this metric that is the safer default — a missing series
is not mistaken for "listed".

## Tests

New `TestMonitorDNSBLTemperror` (root package, same `!windows` build tag
as `serve_unix.go` where the gauge lives): seeds the gauge to `1`, asserts
a temperror leaves it at `1`, then asserts `pass` → `1` and `fail` → `0`
still work. Red without the fix ("gauge is 0 after temperror"), green with
it. `go test ./dnsbl/... ./dns/... .` all pass, `go build ./...`, `go vet`
and `gofmt` clean.

Fixes #445

### DIFF
--- a/dnsbl_monitor_test.go
+++ b/dnsbl_monitor_test.go
@@ -1,0 +1,73 @@
+//go:build !windows
+
+package main
+
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/mjl-/mox/dns"
+	"github.com/mjl-/mox/dnsbl"
+)
+
+// dnsblMetricValue returns the value of the mox_dnsbl_ips_success gauge for the
+// given zone and ip.
+func dnsblMetricValue(t *testing.T, zoneName, ip string) float64 {
+	t.Helper()
+	reg := prometheus.NewRegistry()
+	reg.MustRegister(metricDNSBL)
+	mfs, err := reg.Gather()
+	if err != nil {
+		t.Fatalf("gathering metrics: %v", err)
+	}
+	for _, mf := range mfs {
+		if mf.GetName() != "mox_dnsbl_ips_success" {
+			continue
+		}
+		for _, m := range mf.GetMetric() {
+			labels := map[string]string{}
+			for _, lp := range m.GetLabel() {
+				labels[lp.GetName()] = lp.GetValue()
+			}
+			if labels["zone"] == zoneName && labels["ip"] == ip {
+				return m.GetGauge().GetValue()
+			}
+		}
+	}
+	t.Fatalf("mox_dnsbl_ips_success{zone=%q,ip=%q} not found", zoneName, ip)
+	return 0
+}
+
+func TestMonitorDNSBLTemperror(t *testing.T) {
+	zone := dns.Domain{ASCII: "sbl.spamhaus.org."}
+	ip := "1.2.3.4"
+	zoneName := zone.Name()
+
+	// Set a previous value, as a previous successful check would have.
+	metricDNSBL.WithLabelValues(zoneName, ip).Set(1)
+
+	if got := dnsblMetricValue(t, zoneName, ip); got != 1 {
+		t.Fatalf("initial gauge value is %v, expected 1", got)
+	}
+
+	// A temporary DNS error, e.g. "server misbehaving", is not an indication that
+	// our IP is in the block list, it should not update the gauge.
+	updateDNSBLMetric(zone, ip, dnsbl.StatusTemperr)
+
+	if got := dnsblMetricValue(t, zoneName, ip); got != 1 {
+		t.Fatalf("gauge is %v after temperror, expected it to keep the previous value 1", got)
+	}
+
+	// A successful lookup resets the gauge to 1.
+	updateDNSBLMetric(zone, ip, dnsbl.StatusPass)
+	if got := dnsblMetricValue(t, zoneName, ip); got != 1 {
+		t.Fatalf("gauge is %v after pass, expected 1", got)
+	}
+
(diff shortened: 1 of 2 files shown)


mjl-:
Hi @Huang-404-Q, thanks for the PR!

I think it's a bit verbose. Both the text, and the changes. The could've been just adding an "if" around the lines in serve_unix.go that set the metric. That doesn't need an entire test. This looks LLM generated, I suppose that's the reason?

I'm thinking about these changes:

- I think we shouldn't entirely ignore temporary DNS failures indefinitely. It's not great to not know about blocklisting status if there are temporary dns errors. I think we should still set the metric to "1" if we've been seeing temporary dns errors for X minutes (perhap 60 minutes?).

- We should probably also do the DNSBL checks if there was a temporary error in the previous check. The for-loop currently skips the loop if the previous check was within 3 hours ago and there weren't a lot of outgoing smtp connections.

- We should probably also check the dns resolver error more closely for temporary errors (net.DNSError.Temporary()) before returning StatusTemperr.

Huang-404-Q:
Thanks for the review — you're right on all counts, including the last one. I've cut this down to the guard you suggested:

```go
// A temporary dns error says nothing about whether our ip is in the block
// list, so keep the last known value instead of resetting the metric.
if status != dnsbl.StatusTemperr {
	var v float64
	if status == dnsbl.StatusPass {
		v = 1
	}
	metricDNSBL.WithLabelValues(zone.Name(), ip.String()).Set(v)
}
```

One file, `serve_unix.go`, 8 insertions / 4 deletions. The extracted helper and the whole test file are gone.

I want to be straight about the "looks LLM generated" part, since you asked directly: yes, I used an LLM to write it, and the verbosity is exactly the failure mode you're describing. That's my mistake to correct, not something to argue about. I've read the code above and I'll answer for it.

On the three behaviour points — you're right, and I'd rather do them properly than fold them into this PR:

1. **Indefinitely ignoring temporary errors is wrong.** I'll track when the current temperror streak started per (zone, ip) and set the metric to 1 once it passes a threshold, so a persistent resolver failure still shows up as "unknown/possibly listed" instead of freezing a stale value forever. Is 60 minutes the threshold you want?

2. **Resuming checks after a temporary error** — agreed. I'll make the skip condition also require that the previous check didn't end in a temperror, so a failure doesn't lock the loop out for up to 3 hours.

3. **`net.DNSError.Temporary()`** — agreed, and worth being careful: `IsNotFound` already discriminates the not-listed answer, so anything reaching the error branch is either a real temporary failure or something else. Narrowing to `Temporary()` keeps the "keep last value" branch honest and lets other errors keep their current behaviour.

Do you want 1–3 as a follow-up PR, or would you rather have them in this one?

PR #472, opened 2026-09-09
add LOGIN support to transport auth

### LINKED ISSUE #471: FR: Add support for LOGIN auth method to transports
Transports currently only supports SCRAM-SHA-256-PLUS, SCRAM-SHA-256, SCRAM-SHA-1-PLUS, SCRAM-SHA-1, CRAM-MD5, PLAIN as valid authentication methods. Azure ACS (in Microsoft infinite wisdom) only supports LOGIN.

I'm not sure if there was discussion elsewhere or a particular reason for not supporting LOGIN (aside for it being legacy, though it is supported for Submission)? It seems about as insecure as PLAIN.

blu3id:
Closes #471

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -260,7 +260,7 @@ type TransportSMTP struct {
 type SMTPAuth struct {
 	Username   string
 	Password   string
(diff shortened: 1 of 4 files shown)


mjl-:
thanks for the PR @blu3id!

